PR #13, opened 2023-04-12
Baseline doc for User Consent Management

jordonezlucena:
This PR provides a first draft of the doc for user consent management. The aim is to agree on the ToC.


hdamker:
@jordonezlucena off-topic, just about the contribution process (see [CONTRIBUTING.md](https://github.com/camaraproject/Governance/blob/main/CONTRIBUTING.md)): the official contribution process is "fork and pull" ... which means the preparation of a pull request should happen in a fork of the repo (it's just one click to create it on the top of the repo) and then the pull request should be done from there. Creating a branch within the main repo is an exception you can do as codeowner (with write access to the repo) but no other contributor. The advantage is that you have less risk to accidentially change something in the main repo.

eric-murray (approved):

documentation/user-consent.md
```diff
@@ -0,0 +1,214 @@
...
+
+This document captures guidelines for Operator NaaS platform to handle user consents to comply with GDPR or equivalent requirements in an easy way, introducing the concept of purpose in OpenID Connect. Even being defined based on concepts that maps to GDPR regulation, proposed solution and concepts are generic enough to be used by Operators on any country.
+
+>[TO BE EDITED/COMPLETED]
+
+## Glossary of Terms and Concepts
+
+The list below introduces several key concepts:
+
+-	`NaaS`: Network-as-a-Service model where Telco Network capabilities are exposed to third parties through APIs. And some of those APIs are likely to require user consent to be accessed.
```
shilpa-padgaonkar, line 42: Submitted an issue already for this https://github.com/camaraproject/IdentityAndConsentManagement/issues/17

documentation/user-consent.md
```diff
@@ -0,0 +1,214 @@
+# User Consent Management
+
+This document captures guidelines for Operator NaaS platform to handle user consents.
+
```
shilpa-padgaonkar, line 4: Outlining what managing user consent means would be helpful.
Some points that can be considered
- Select an identifier which can be used to create the user consent records
- Allow the operator to create, read, update, and delete consent data records 
- Validate an API request against a policy enforcement point to check if consent already exists or not
- Allow users to view and update their consents 
- Describe a flow how consent can be collected from the user without degrading the user experience while using a third-party service

documentation/user-consent.md
diegogonmar: Fine, I removed it in this commit https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/150a596f77633aa771dbadbebf6bc00b107a5762

documentation/user-consent.md
diegogonmar: I included an explanation in introduction chapter, in commit https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/c7b23f53b47ba92afa9eab8aeb36083962b76cfe. The document will then get into details, in different PRs as this one is intended to set the baseline.

I included your suggestions with some rewording, please check.

documentation/user-consent.md
```diff
@@ -0,0 +1,220 @@
...
+Depending on the purpose categories the purpose may have a specific level that determines the processing allowed by the user if the purpose is granted. **User consent is required to gain access to non-automatic purposes**.
+
+| **Purpose Level** | Description | Category |
+| --- | --- | --- |
+|`legal_obligation` | A process that is required by law. | Automatic purpose |
+|`contract` | A process that is needed to fulfill the service contract with the user, and it is already approved when the contract was signed. | Automatic purpose |
+|`legitimate_interest`| An interest of the data processor that does not conflict with the fundamental interest of the individuals. | Opt-out purpose |
+|`compatible` | A re-processing of data acquired through another purpose that is compatible with the first purpose. | Opt-out purpose|
+|`consent` | This level applies to purposes which are unrelated enough to the service. | Opt-in purpose|
+|`terms_and_conditions`| The operations or tasks done by an application to provide its functionality, as described in the Terms and Conditions. | Opt-in purpose|
```
rartych, line 91: Are there any reference for proposed **Purpose levels**?
If the levels are ordered from the most strict to less significant, then the `consent` level should be the last one, I guess.

The description for `compatible` level is not clear - what is *another purpose* and what is *first purpose*.
The name for `consent` level is ambiguous since **User consent is required to gain access to all non-automatic purposes**. Can it be called `voluntary` or `optional` or other?

documentation/user-consent.md
```diff
@@ -0,0 +1,220 @@
...
+      - [CIBA flow](#ciba-flow)
+      - [Client Credentials flow](#client-credentials-flow)
+      - [... (other grant types)](#-other-grant-types)
+  - [Consent API](#consent-api)
+  - [Annex: Flow source code](#annex-flow-source-code)
+
+
+## Introduction
+
+Operator platform implementing CAMARA should be built with a privacy-by-design approach to fully comply with data protection regulations like the [GDPR regulation](https://gdpr-info.eu/) in Europe, which provides a great level of protection for user's privacy. This means that an API that processes Personal Information needs a user consent. Consents are given by users to legal entities to process personal data under a specific purpose.
```
chrishowell, line 32: this isn't true - an API that processes Personal Information needs a 'legal basis' for doing so, one of which would be user consent.

documentation/user-consent.md
```diff
@@ -0,0 +1,220 @@
...
+
+This document captures guidelines for Operator platform to handle user consents to comply with GDPR or equivalent requirements in an easy way, introducing the concept of purpose in OpenID Connect. Even being defined based on concepts that maps to GDPR regulation, proposed solution and concepts are generic enough to be used by Operators on any country.
```
chrishowell, line 34: Why are we introducing the technical solution, e.g. OpenID Connect within the introduction. IMO we should keep technical solutions out of the 'baseline'

documentation/user-consent.md
```diff
@@ -0,0 +1,220 @@
...
+>[TO BE EDITED/COMPLETED]
+
+## Glossary of Terms and Concepts
+
+The list below introduces several key concepts:
+
+-	`Application`: client system that requires access to protected resources. Application must use the appropriate access token to access those resources (e.g. CAMARA Network APIs).
+-	`User`: the human participant which is identified in Telco Operator by a unique user identifier (e.g. Subject identifier sub in OpenID Connect terminology). The user is the resource owner.
+-	`Auth Server`: authorization server which receives requests from applications to issue an access token upon successful authentication and consent of the user. The OpenID Connect Provider is able to authenticate the user validating user identity against the corresponding Identity Provider. The authorization server exposes two endpoints: the Authorization endpoint and the Token endpoint.
+-	`Identity Provider (IdP)`: It corresponds to the OpenID identity provider which is the party that provides user authentication as a service (it creates, maintains, and manages user identity information).
```
chrishowell, line 52: This doesn't need to be OpenID Connect does it? Again, let's remove technology from the definition.

documentation/user-consent.md
jpengar: @shilpa-padgaonkar Changes included are ok with you?

documentation/user-consent.md
jpengar: Yes, I agree. We will update the document to rephrase it correctly. As explained above, we were using user consent as a generic term and not specifically linked to the "consent" legal basis. But it does seem to be causing some confusion, so we will try to change the doc accordingly.
If you have any suggestions to do so, please don't hesitate to suggest a change.

documentation/user-consent.md
jpengar: During the WG kick-off, it was proposed to define a solution following the OIDC standard (also in line with the previous discussion we had in the Communities WG on this topic). Therefore it is included in the baseline document. But of course, if you have another suggestion, feel free to put it up for discussion.

documentation/user-consent.md
jpengar: See https://github.com/camaraproject/IdentityAndConsentManagement/pull/13#discussion_r1195033670

documentation/user-consent.md
jpengar: Will include this changes if it is ok with you @rartych

jpengar: Fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/4b761353f33ee263e98e8309c8fb698908401206

documentation/user-consent.md
jpengar: Fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/4b761353f33ee263e98e8309c8fb698908401206

documentation/user-consent.md
jpengar: Further clarifications included in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/4b761353f33ee263e98e8309c8fb698908401206

documentation/user-consent.md
```diff
@@ -12,7 +12,7 @@ This document captures guidelines for Operator platform to handle user consents.
     - [Purpose levels](#purpose-levels)
     - [Purpose relationship with other key concepts](#purpose-relationship-with-other-key-concepts)
     - [Purpose definition](#purpose-definition)
-    - [Using purpose parameter in the authorization request](#using-purpose-parameter-in-the-authorization-request)
+    - [Applying purpose concenpt in the authorization request](#applying-purpose-concenpt-in-the-authorization-request)
```
gmuratk, line 15: "concenpt" --> "concept" in 2 places

documentation/user-consent.md
```diff
@@ -29,11 +29,12 @@ This document captures guidelines for Operator platform to handle user consents.
...
 
-This document captures guidelines for Operator platform to handle user consents to comply with GDPR or equivalent requirements in an easy way, introducing the concept of purpose in OpenID Connect. Even being defined based on concepts that maps to GDPR regulation, proposed solution and concepts are generic enough to be used by Operators on any country.
+As per CAMARA Commonalities ["Authentication and Authorization Concept for Service APIs"](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/CAMARA-AuthN-AuthZ-Concept.md), CAMARA API access will be securitized using [OpenID Connect](https://openid.net/connect/) on top of OAuth 2.0 protocol. This document captures guidelines for Operator Exposure platform to manage CAMARA API access and user consent to comply with GDPR or equivalent requirements in an easy way, introducing the concept of purpose in OpenID Connect. Even being defined based on concepts that maps to GDPR regulation, proposed solution and concepts are generic enough to be used by Operators on any country.
```
gmuratk, line 34: Should it be "securitized" or "secured"?

gmuratk, line 34: To match with earlier change "This document _defines_ guidelines..."

documentation/user-consent.md
jpengar: Thanks. Fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/5ab70dc278f926455ece38c1294586ed1c84dfa6

documentation/user-consent.md
jpengar: Fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/0162b0ae7a5045d187219ad55f0dd2a2dd45f9d1

jpengar:
> Hi,
> 
> regarding the concept of purpose, we (Orange) would like to correctly understand the benefits to add the purposes to the OIDC standard. We already have a group component using the standard and there will be impacts in this case. Our understanding is that if purpose is studied/proposed, it means that the standard with scopes isn't sufficient for all the use case needed in the camara perimeter. However, we currently don't know which use case won't be covered by the standard. Or is it just in order to "provides a human-friendly description" ? We also wonder if adding the purpose would complicate the standard with another layer to define, administer, manage changes, agree on vocabulary etc.
> 
> thank you in advance for the clarifications

@sebdewet This is actually a really good point and is open for discussion. I've created a specific github issue (https://github.com/camaraproject/IdentityAndConsentManagement/issues/32) to get agreement on this. Because I think we can discuss it separately without interfering with this PR, which is just intended to merge a first baseline template document proposal that will need to be progressively edited and developed in the near future to document the complete final agreed solution.

jpengar:
Reviewers (@chrishowell @gmuratk @shilpa-padgaonkar @sebdewet @rartych @eric-murray) and all, do you agree (given the latest changes included in the template document proposal) to merge the document as a starting point for discussion?
As we said in the last call, everything in the document so far is set in stone and the document will evolve and everything can be discussed/changed. I don't see the point in being too strict about the content of the document, as it is only a draft template proposal.

hdamker:
@jpengar 
> everything in the document so far is set in stone
Guess you meant the opposite, right? 😃

jpengar:
> @jpengar 
> 
> > everything in the document so far is set in stone
> 
> 
> 
> Guess you meant the opposite, right? 😃 

Edited. XD. Exactly, just the opposite... My bad.

documentation/user-consent.md
hdamker: On behalf of @shilpa-padgaonkar: yes, changes ok.

gmuratk (approved):

documentation/user-consent.md
```diff
@@ -0,0 +1,169 @@
...
+
+>[TO BE EDITED/COMPLETED - [OAuth 2.0 Authorization Framework](https://www.rfc-editor.org/rfc/rfc6749)]
+
+## User Identity
+
+>[TO BE EDITED/COMPLETED] - User Identification: Get an identifier of a user by some piece of information. E.g.: IP, phone number, document number...
+
+## User Authentication/Authorization & Consent Management
+
+>[TO BE EDITED/COMPLETED] - The base standard is OpenID Connect. User authentication: Verify that the user is who they say they are. For example: IP/Network Authentication, SMS OTP, User/Password, Push to Operator App... it depends on the type of authorisation (grant type) chosen according to the application use cases, integration model and requirements (B2B/B2C, via aggregator or not, delegated consent or not...). User authorisation: Verify access to the requested resource (OpenID Connect scope, purpose) through policies and rules... user consent is checked at this point as a pre-requisite before issuing an access token.
```
hdamker, line 129: "Verify that the user is who they say they are" ... that's correct definition of authentication, but confusing together with the above definition of `User` as the subject (who is not verified ... at least only indirectly). Maybe we should use consequently define "End User" to distinct from the API/capability user/consumer which will be authenticated.

documentation/user-consent.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Authentication mechanism/s
+
+>[TO BE EDITED/COMPLETED] - Authentication mechanism/s to be considered... E.g.: IP/Network Authentication, SMS OTP, User/password, Push to Operator App​...
+
+### Consent capture and storage
+
+>[TO BE EDITED/COMPLETED] - Consent capture (by Operator vs by API Client a.k.a delegated) and storage (in Operator vs in a third party).
+
+### Athorization flows / grant types
```
hdamker, line 139:
```suggestion
### Authorization flows / grant types
```

hdamker (commented):
Document template is fine for me. Small edit comments above. 

My general concern is the tight coupling of the OIDC authorization flow for the API consumer with the consent check towards the subject. The pro is definitely that with the access token also the consent is given, but it could be also a cons as the existence of this token already reveals that there is consent given (which in itself is PII).

We should carefully check if this will fit for all scenarios (e.g. event subscriptions for a set of devices or were consent need to be granted upfront to avoid an interruption of the actual service). Another scenario could be that the API/capability consumer does not have a sufficient identifier beforehand, but only some data which will be resolved on operator side into an identifier ... read: a scenario were the consent can only be checked within the service logic and the API consumer might not need to get an identifier of the subject.

documentation/user-consent.md
```diff
@@ -0,0 +1,169 @@
...
+- Flows detailing how Consent can be collected from the User without degrading the user experience while using a third-party service in an Application.
+
+>[TO BE EDITED/COMPLETED]
+
+## Glossary of Terms and Concepts
+
+The list below introduces several key concepts:
+
+-	`Application`: client system that requires access to protected resources. Application must use the appropriate access token to access those resources (e.g. CAMARA Network APIs).
+-	`User`: the human participant which is identified in Telco Operator by a unique user identifier (e.g. Subject identifier sub in OpenID Connect terminology). The user is the resource owner.
```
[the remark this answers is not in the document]
jpengar: Fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/5cb8ed958e0cba62bc76a8c610b7679bba3016ff normalizing terms to lower case.

documentation/user-consent.md
```diff
@@ -0,0 +1,169 @@
...
+-	`Auth Server`: authorization server which receives requests from applications to issue an access token upon successful authentication and consent of the user. The OpenID Connect Provider is able to authenticate the user validating user identity against the corresponding Identity Provider. The authorization server exposes two endpoints: the Authorization endpoint and the Token endpoint.
```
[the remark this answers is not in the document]
jpengar: Fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/5cb8ed958e0cba62bc76a8c610b7679bba3016ff

documentation/user-consent.md
jpengar: fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/5cb8ed958e0cba62bc76a8c610b7679bba3016ff

documentation/user-consent.md
```diff
@@ -0,0 +1,169 @@
...
+
+- Scopes tell what resources (e.g. APIs) are accessible under those scopes.
+- A purpose has associated a set of personal information scopes (PI scopes). 
+  - These are scopes that refer to resources providing access to Personal Information and thus requiring special protection as stated by e.g. the GDPR law.
+  - Getting access to PI scopes MUST always be done by explicitly declaring a purpose.
+- Purposes are assigned to an application.
+
+Therefore, an application will have access to:
+  - APIs or resources providing personal information under the assigned purpose (because of the PI scopes included in that purpose)
+    - **Only** if user has granted her consent for that purpose
```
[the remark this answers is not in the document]
jpengar: Fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/13/commits/5cb8ed958e0cba62bc76a8c610b7679bba3016ff

documentation/user-consent.md
jpengar: That is just a note to be removed with the actual content of that section. It shouldn't be a stopper to merge the template proposal.

In any case, from our point of view on what refers to CAMARA APIS, the end user is the human participant which is identified in telco operator by a unique user identifier (e.g. subject identifier [as defined by OIDC](https://openid.net/specs/openid-connect-core-1_0.html)). Therefore, the Authentication process allows to confirm/validate user identity. We are referring as user as the person who uses the operator capabilities (end user of the device). But normally they will be the person who has a contract relationship with the telco operator i.e. the MNO subscriber.
 
Just to provide an example, in case of Telefónica in Brazil, there are family plans with multiple phone numbers. One "owner" phone number (the one corresponding to the user having the contract relationship) and other dependent phone numbers (used by other family members not being the owner). And we have the mechanisms in place to have different users for each phone number (different user id and profile). Otherwise, from Operator perspective the end user will be always the MNO subscriber.

jpengar:
> We should carefully check if this will fit for all scenarios (e.g. event subscriptions for a set of devices or were consent need to be granted upfront to avoid an interruption of the actual service). Another scenario could be that the API/capability consumer does not have a sufficient identifier beforehand, but only some data which will be resolved on operator side into an identifier ... read: a scenario were the consent can only be checked within the service logic and the API consumer might not need to get an identifier of the subject.

This is a really good point, and there is a specific issue (https://github.com/camaraproject/IdentityAndConsentManagement/issues/21) where all the necessary scenarios to cover should be discussed and agreed upon. Including the ones you mention. We can move the discussion there if you like.

jpengar:
@gmuratk Since you approved the content of this PR, should we close the baseline document feedback issues you opened? https://github.com/camaraproject/IdentityAndConsentManagement/issues/25, https://github.com/camaraproject/IdentityAndConsentManagement/issues/26, https://github.com/camaraproject/IdentityAndConsentManagement/issues/27 & https://github.com/camaraproject/IdentityAndConsentManagement/issues/28

documentation/user-consent.md
hdamker: Fully agree that this is not a stopper to merge. It was just a proposal to distinct API user/consumer (= application) more clearly form end user. You don't need to consider this comment here.

hdamker:
> I don't really see the problem here.

@jpengar let's postpone this discussion until we have more concrete flows on the table.

gmuratk:
> @gmuratk Since you approved the content of this PR, should we close the baseline document feedback issues you opened? #25, #26, #27 & #28

I closed these issues as they are addressed with the changes introduced in the recent commit.

jpengar:
Could you take a look and give your final approval? I think we should be able to merge the baseline document now...

shilpa-padgaonkar (approved):
Thanks for the initial doc proposal. Fine to approve the base template with the assumption that the document needs further detailing and changes going ahead.
For purpose, a new issue being created is appreciated as we are not sure if adding a new header would be the best approach. 
Also, consent by the end user will be given against a privacy policy that has been agreed between the API consumer and the operator, rather than just a simplistic purpose.

jpengar (approved):

rartych (approved):

PR #23, opened 2023-05-10
Add changelog and gitignore file

ShutingQing:
#### What type of PR is this?
* documentation

1. Problem: 
Currently, we don't have a version control and changelog sys set up. So in the future, it's hardly can someone tell which part was did in v1, and which part was did in v2 or v3. 

Solution: 
So I created a changelog file. Each time someone commits, can leave a notes in changelog. Changes and versions are supposed to be included. 

3. Problem: 
When we commit, we have to exclude sys file/folder generated by IDE, like I'm using IDEA, there will always be a .idea folder. So as other IDE. 

Solution: 
I added a gitignore file. People can add your IDE generated file/folder here. So each time when you commit, you do not have to exclude those folders before you do a git push. 

5. Added some other complimentary documentations

Linked to [https://github.com/camaraproject/IdentityAndConsentManagement/issues/24](url)

#### Changelog input

```
 release-note

v0.1.0 wip documentation updates

#### Additional documentation 

This section can be blank.



```
docs

```

### CODE REVIEW

**Comment by hdamker** [comment]
Hi @ShutingQing, great to do some housekeeping here.

Just two remarks:
* There is no such thing as a 0.1.0 release yet, as there is no API definition as well
* documentation/API_documentation/README.MD is not meant to hold the documentation but it is just a placeholder in the directory until there is the first documentation file will be placed there (i would also propose not to mix housekeeping and documentation content within one PR)

**Comment by ShutingQing** [comment]
So identity and consent management will not do some common api stuff for user authentification or consent, like this is only a repo for people to reach a agreement?

**Comment by hdamker** [comment]
> So identity and consent management will not do some common api stuff for user authentification or consent, like this is only a repo for people to reach a agreement?

That's not what i meant ... actually the discussion was moved into an own repository as there will be most probably "api stuff". But currently there isn't anything. I would consider the current state "work in progress" with the goal to come to v0.1.0 release eventually (using v0.1.0 already now could give the false impression that the content in `main` is already released).

**Comment by ShutingQing** [comment]
> ith the goal to

Oh yeah, got what you mean.

**Comment by jpengar** [comment]
> That's not what i meant ... actually the discussion was moved into an own repository as there will be most probably "api stuff". But currently there isn't anything. I would consider the current state "work in progress" with the goal to come to v0.1.0 release eventually (using v0.1.0 already now could give the false impression that the content in main is already released).

I fully agree. We should follow the guidelines set by Commonalities on this matter: https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/SupportingDocuments/Camara_Versioning_Guidelines.md

**Comment by ShutingQing** [comment]
> > That's not what i meant ... actually the discussion was moved into an own repository as there will be most probably "api stuff". But currently there isn't anything. I would consider the current state "work in progress" with the goal to come to v0.1.0 release eventually (using v0.1.0 already now could give the false impression that the content in main is already released).
> 
> I fully agree. We should follow the guidelines set by Commonalities on this matter: https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/SupportingDocuments/Camara_Versioning_Guidelines.md

Yes. Have marked the version to wip. So that's why I added a changelog file. As the guidelines states, our sub project needs to record main changes, including features, fixes and so on. 

Or that when it comes to version v0.3.0, people hardly can not what are done in v0.1.0, what are done in v0.2.0.

**Comment by rartych** [comment]
Current Github PR template includes section: 
#### Changelog input

```
 release-note

```
It is intended to provide short description of the PR that will be aggregated in the changelog when the release is published.
So instead of updating changelog after each PR merge, we assume that the needed content is provided in PRs and can be used to describe the release.
Since we are in the early stage of the subproject as discussed above, I propose to keep the just the template of  CHANGELOG.md file.

**Comment by ShutingQing** [comment]
Okay, sure.

**Comment by ShutingQing** [comment]
I have replaced the changelog file to template. And remove the placehoder api documentation readme. Pls review.

File: documentation/MeetingMinutes/README.MD [hunk 1]
```diff
@@ -1 +1,30 @@
...
+| 2023/08/23 |
+
+
+### MoMs
+
+2023/03/22 [Notes](MOM-2023-03-22.md)
+
+2023/04/12 [Notes](MOM-2023-03-22.md)
+
+2023/05/03 [Notes](MOM-2023-05-03.md)
```
jpengar, line 30: This will require us to update this README every time we add a meeting minutes document.

jpengar (approved):
Generally speaking, LGTM. The only thing we need to check is whether we want to add an entry in MeetingMinutes README.MD every time we add a new Meeting Minutes document...

documentation/MeetingMinutes/README.MD
ShutingQing: looks good, thanks.

Issue #21, opened 2023-04-28
Definition of scenarios for consent management 


rartych:
**Problem description**
The need for Consent Management API is widely accepted. However, there are different approaches and requirements for this API.
There are also multiple scenarios for consent management.

**Expected action**
I propose to start with definition of scenarios for user consent use and management in CAMARA, that would help to elicit requirements and define API specification. It can be done in  additional section of the proposed baseline document or in a separate document  (such as User Story within documentation in some subprojects). The scenarios can start from simple offline granting and revoking consent by the user and online checking consent with CAMARA Service API call, and include more complex one when the consent is given just before CAMARA Service API call.
Scenarios should present actions and roles of all actors involved.
Description of high-level scenarios (without going into details of protocol flows) should also help legal teams analyze the proposed approach for consent management.

**Additional context**
Scenarios can be illustrated with Sequence Diagrams that can be created  using [Mermaid tool ](https://mermaid.js.org/) and [included directly in Github Markdown files](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/).

jpengar:
PR #33 has been restored in https://github.com/camaraproject/IdentityAndConsentManagement/pull/74

PR #33, opened 2023-05-31
Consent-related use cases 

bigludo7:
Draft version - List of Use Case illustrating (or not) requirement for consent.

#### What type of PR is this?

* documentation

The objective of the PR is to contribute to a document listing consent related use-cases. Thi document will be use as an output for Consent API work.

No issue

This is a **DRAFT** document. 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/MeetingMinutes/MOM-2023-09-06.md
+++ b/documentation/MeetingMinutes/MOM-2023-09-06.md
@@ -1,0 +1,101 @@
+# Identity and Consent Management meeting
+
+#### *6th of September 2023*
+
+<br>
+
+## Attendees
+
+
+| Companies | Attendees |
+| --------- | --------- |
+| Telefónica | Guido García, Diego Gonzalez, Jose Antonio Ordoñez |
+| Vodafone  | Sebastian Köller |
+| Deutsche Telekom AG | Rafal Artych, Axel Nennker, Dawid Wróblewski |
+| Orange    | Robert Ludovic, Sebastian Dewet |
+| GSMA      | Daniel Okonkwo, Tom van Pelt, Toyeeb Rehman |
+| T-Mobile US | Murat Karabulut |
+| Verizon   | Syed A Rehman |
+
+<br>
+
+## Agenda
+
+1. Recent updates & recap
+2. Issues and PRs: Priority discussions:
+    - Discussion on Issue #32. "The concept of purpose".
+3. Action Points 
+4. AoB
+
+<br>
+
+## Recent updates & recap
+
+Telefonica presents the status of WG:
+
+- Issues closed since last call 23/8
+  - Issue #15 "Subscriber Identity API".
+  - Issue #52 "Differences in authentication on APIs between deployments can lead to differences in API usage and break federation".
+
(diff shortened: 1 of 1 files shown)


rartych:
Fixes #21

shilpa-padgaonkar:
@bigludo7 : Thanks for the PR. 
Good to see different usecases elaborated for individuals vs enterprise, data subjects versus no-data subject and onetime vs time-limited consent.

We have thought of certain scenarios (though I have not explicitly written them in user story format). 

Just listing them down below for you to take a look. (with No expectation that any part makes its way into the doc in some form :) )

**Scenario1:** Target device, user device, authentication device are the same.
Can be safely assumed that we can request consent from the subscriber/end user

**Scenario2:** Target device and user device are the same but authentication device is different. For e.g. Family sim card

**Scenario3:** User device and authentication device are same but target device is different. For e.g. Parent using tracking app for child’s device

**Scenario4:** Target device, user device  and authentication device are all different. 

Two main aspects need to be considered in scenario 2, 3 and 4:
·       Target device could be used by a minor, adult (family sim cards), employee (enterprise mobile devices -- this seems to be already covered in one of your user stories) or does not target a data subject at all (covered in IoT case user story but only for enterprise not individual).
·       Authentication device could have a sim from the same or from a different MNO.

rartych:
Since in the consent document we have the definition of :
_`Aggregator`: aggregate Operator’s CAMARA standardised APIs for building services offered to application developers. An aggregator can be a **hyperscaler** offering its own services or directly exposing CAMARA APIs available at the operators, or it can be a telco operator acting as an aggregator_

I propose to replace _hyperscaler_ with _aggregator_ in this document.
@bigludo7 is it OK for you if I prepare change suggestions that you can commit in the bulk?

bigludo7:
@rartych Sure ! please move forward. I need also to take into account Shilpa comments.

rartych:
Another interesting use case is considered in [Blockchain Public Address](https://github.com/camaraproject/BlockchainPublicAddress/issues/16) subproject.

Currently it considers giving consent to data access not to applications/service providers but to specific subscribers identified by MSISDN using this applications - kind of whitelisting who can get information about given subscriber from Telco operator.

The problem may be also solved on API level (not using consent management system) or even on the specific application level.

This problem can be also interesting for identity related use-cases.

gmuratk:
I recall a document with a few use cases listed, but I can't seem to locate it at this time.  
In the meeting notes from Sept 6, 2023, it is stated that discussion and progress will continue offline. Is it possible to gain visibility into progress?  
Also Is this progress expected to produce a new section in the [https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#introduction](url)?  
Thank you,

bigludo7:
@gmuratk Yes my bad - I've accidentally remove this file. I will reinsert it.

jpengar:
This PR has been restored in https://github.com/camaraproject/IdentityAndConsentManagement/pull/74

PR #55, opened 2023-08-09
Include network-based auth info in authN-AuthZ doc

### LINKED ISSUE #45: Add missing details in authN-authZ doc
**Problem description**
This issue https://github.com/camaraproject/NumberVerification/issues/46 resolution should be piggy backed into the AuthN-AuthZ document 

**Expected action**
Extend AuthN-AuthZ doc with the details about network based authN

**Additional context**

jpengar:
#### What type of PR is this?

* documentation

Extend AuthN-AuthZ doc with the details about network based authN

Fixes #45

N/A

#### Changelog input

N/A

#### Additional documentation 

N/A

### DIFF
--- a/documentation/CAMARA-AuthN-AuthZ-Concept.md
+++ b/documentation/CAMARA-AuthN-AuthZ-Concept.md
@@ -136,6 +136,15 @@ It enables Clients to verify the identity of the End-User based on the authentic
 <img src="./images/oidc.png" alt="OIDC"
 	title="Abstarct OIDC Flow" width="350" height="300" />	
 
+ OIDC provides a framework for identity providers to support a wide range of authentication methods/factors that can be used to verify the identity of the user:
+
+ - Username and Password: The user provides the client application with a username and password. The client application then sends the username and password to the identity provider, which verifies the user's identity and returns an ID token to the client application.
+ - SMS One-Time Password (OTP): The user provides the client application with a phone number. The client application then sends the phone number to the identity provider, which sends an SMS message to the user's phone number. The user enters the code from the SMS message into the client application, which sends the code to the identity provider. The identity provider verifies the code and returns an ID token to the client application.
+- Network-based authentication: An authentication mechanism based on the identification of the endpoint of a network connection (e.g., by the observed IP address of the user device connected to the network). Network operators know which user is associated with a network resource at any given time, such as the mobile phone number associated with a particular cellular connection.<br>This is particularly interesting in the context of CAMARA APIs exposed by network operators. For example, it is the authentication method defined for the [Number Verification API](https://github.com/camaraproject/NumberVerification) without any user interaction in the process (aka "silent" auth).
+- Multi-factor authentication (MFA): MFA combines two or more authentication factors to increase security. For example, a combination of a password and a push notification to a registered mobile device. MFA is often used to provide an extra layer of security for sensitive applications such as banking.
+
+In fact, there are many more authentication methods/factors that can be used with OIDC, which can have different levels of security based on the factors they rely on to verify a user's identity (e.g., biometrics, hardware security keys, social identity providers such as Google, Facebook, etc.).
+
 
 #### Authorization code grant
 The authorization code is a temporary code that the client will exchange for an access token. The code is obtained from the authorization server. The user gets to know about the information requested by the client and can then approve or deny the request. The application requesting for the access token must supplement the request with the client secret. This reduces the risk that the authorization code can be exploited if intercepted by a bad actor.


sebdewet:
Hi,

> On what refers to CAMARA APIs to be exposed by Telco operators, the end user is the human participant which is identified in Telco Operator by a unique user identifier

> It is important to differentiate the identification of a user from the identification of a device or UE. For example, the same user could have multiple associated devices.

A user can also have multiple associated account. It means that several identifiers can correspond to a unique user.

> line 29 : ### User indentity

change with identity

jpengar:
Thanks for your comments @sebdewet, I don't think they're related to the content of this PR, but I've tried to address them below anyway.

>Hi,
>A user can also have multiple associated account. It means that several identifiers can correspond to a unique user.

Commit https://github.com/camaraproject/IdentityAndConsentManagement/pull/55/commits/42dc095ec1b21594ecb64fd84cf3a167e3bd90e1 will hopefully fix it. I've also added further clarification to be consistent with what is defined in the `user-consent.md` glossary. 

>line 29 : ### User indentity
>change with identity

As for the other suggestion, I would rather keep "User identity".

sebdewet:
Hi,

My bad, I wanted to change "User indentity" by "User identity". I should have precised correctly.

It's just that there is useless "n"

sebdewet (approved):

rartych (dismissed):

AxelNennker (changes requested):
1) Please add a reference to CIBA
https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html

2) Please fix a typo: Abstract, also make the alt text "Abstract OIDC flow"
OIDC	

3) Please add sim-based authentication aka silent-auth2
SIM-based authentication: a privileged mobile application can perform EAP-AKA and authenticate the user (subscriber, sim identity). The operator token retrieved from the operator after EAP-AKA can be added to the CAMARA API request as a user identifier. SIM-based authentication is similar to network-based authentication as it leverages unique operator capabilities.

diegogonmar:
> 1. Please add a reference to CIBA
>    https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html
> 2. Please fix a typo: Abstract, also make the alt text "Abstract OIDC flow"
>    OIDC
> 3. Please add sim-based authentication aka silent-auth2
>    SIM-based authentication: a privileged mobile application can perform EAP-AKA and authenticate the user (subscriber, sim identity). The operator token retrieved from the operator after EAP-AKA can be added to the CAMARA API request as a user identifier. SIM-based authentication is similar to network-based authentication as it leverages unique operator capabilities.

Hi @AxelNennker. Fine with the first two points, will send a commit. Regarding the third one, thats something that has not been discussed so not sure we can just add it, because this PR is about network authentication and that would be a different thing. 
It also has some complexity and would require specific flow, not only about the part of retrieving the operator token but for the part of using it and operator validating it. It should not be used in CAMARA API request but instead in the request to obtain the Access Token. I suggest you open a separate issue so we can properly discuss, its an interesting topic.

documentation/CAMARA-AuthN-AuthZ-Concept.md
```diff
@@ -136,6 +136,15 @@ It enables Clients to verify the identity of the End-User based on the authentic
...
+ - SMS One-Time Password (OTP): The user provides the client application with a phone number. The client application then sends the phone number to the identity provider, which sends an SMS message to the user's phone number. The user enters the code from the SMS message into the client application, which sends the code to the identity provider. The identity provider verifies the code and returns an ID token to the client application.
+- Network-based authentication: An authentication mechanism based on the identification of the endpoint of a network connection (e.g., by the observed IP address of the user device connected to the network). Network operators know which user is associated with a network resource at any given time, such as the mobile phone number associated with a particular cellular connection.<br>This is particularly interesting in the context of CAMARA APIs exposed by network operators. For example, it is the authentication method defined for the [Number Verification API](https://github.com/camaraproject/NumberVerification) without any user interaction in the process (aka "silent" auth).
```
DT-DawidWroblewski, line 143: Hi! I have some questions to this one:
1. How do developers know what network auth solution works in particular MNO?
2. How does CAMARA want to bring this knowledge and technical flow to this complexity? (I can name at least 4 solutions inside DT... not mentioning how it works worldwide...)

documentation/CAMARA-AuthN-AuthZ-Concept.md
```diff
@@ -136,6 +136,15 @@ It enables Clients to verify the identity of the End-User based on the authentic
...
+- Multi-factor authentication (MFA): MFA combines two or more authentication factors to increase security. For example, a combination of a password and a push notification to a registered mobile device. MFA is often used to provide an extra layer of security for sensitive applications such as banking.
+
+In fact, there are many more authentication methods/factors that can be used with OIDC, which can have different levels of security based on the factors they rely on to verify a user's identity (e.g., biometrics, hardware security keys, social identity providers such as Google, Facebook, etc.).
```
chrishowell, line 146: I think this needs to be standardised for each scope across the CSPs.

documentation/CAMARA-AuthN-AuthZ-Concept.md
diegogonmar: Hi @DT-DawidWroblewski. 
1. No need to know as developer just needs network auth to happen. They trigger the standard AuthCode flow, which is based on http redirects and network auth happens there. It could be auth based on observed public IP or on other mechanisms (notice definition includes use of observed IP as an example, not as the unique valid solution)

2. Why do you think CAMARA should standardize this? It does not matter if e.g. Telefonica uses mapping of public IP to MSISDN while DT uses another mechanism, as long as it fits the concept defined as Network-based authentication as defined above. In Telefonica it similar to DT, we don't have the same exact mechanism in every country

documentation/CAMARA-AuthN-AuthZ-Concept.md
diegogonmar: You refer to standardize the list of available authentication methods? I would say that methods should work the same way, and in CAMARA maybe we may include list of valid methods, once requirements based on use cases appear.

documentation/CAMARA-AuthN-AuthZ-Concept.md
DT-DawidWroblewski: Hi @diegogonmar

1. As long as customer is aware what is going on behind /authorize - it is fine. The only concern I have is that statement **Network-based authentication** might be confusing, e.g. when one MNO use HE and it requires developer to send a request over HTTP (HE does not with HTTPS; how it would work for Web and solutions like *Apple Private Relay*?), another MNO use *redirect* mechanism, so it requires developer to implement a proper software code to handle redirects and on top of this, redirects might use different flows behind, e.g. simple request response or oidc/oauth solutions that involves *code grant* flow. Developer needs to know how to handle particular solution, otherwise she or he gets stuck on proper handling of **Network-based authentication**. 
2. Camara doesn't need to specify this and I think that it should deliver the most valuable information about Telco APIs back to developer.

diegogonmar:
> > 1. Please add a reference to CIBA
> >    https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html
> > 2. Please fix a typo: Abstract, also make the alt text "Abstract OIDC flow"
> >    OIDC
> > 3. Please add sim-based authentication aka silent-auth2
> >    SIM-based authentication: a privileged mobile application can perform EAP-AKA and authenticate the user (subscriber, sim identity). The operator token retrieved from the operator after EAP-AKA can be added to the CAMARA API request as a user identifier. SIM-based authentication is similar to network-based authentication as it leverages unique operator capabilities.
> 
> Hi @AxelNennker. Fine with the first two points, will send a commit. Regarding the third one, thats something that has not been discussed so not sure we can just add it, because this PR is about network authentication and that would be a different thing. It also has some complexity and would require specific flow, not only about the part of retrieving the operator token but for the part of using it and operator validating it. It should not be used in CAMARA API request but instead in the request to obtain the Access Token. I suggest you open a separate issue so we can properly discuss, its an interesting topic.

@AxelNennker points 1 and 2 fixed here https://github.com/camaraproject/IdentityAndConsentManagement/pull/55/commits/4a7ec36d69f052031896f7a8136b61ca62084779

DT-DawidWroblewski:
I think that **Technical ruleset for the Frontend flow** description in :
[CAMARA-API-access-and-user-consent.md](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#user-identity)

might confuse reader.

>The device application (front-end) must be able to handle browser redirects.

this is too specific, as HE solutions do not rely on redirects

>Network based authentication.
Use network based authentication mechanism to obtain the user identifier, i.e.: MSISDN. Set the OAuth sub to the unique user ID in the operator.

this is one way of doing it, but it is not common to all MNOs. I suggest to replace _i.e._ with _e.g._. Usually MNOs avoid sharing MSISDN in a flow, but MSISDN is a result of identifiers exchanged within a flow.

documentation/CAMARA-AuthN-AuthZ-Concept.md
diegogonmar: Hi @DT-DawidWroblewski,
Regarding 1:
- As I indicated before it will be transparent for developer. OAuth2 AuthCode flow is based on HTTP requests and involves redirections. Need to handle redirects is not a requirement or something to be known, its something basic attached to the fact of using Authorization Code. OAuth RFC says: _This specification makes extensive use of HTTP redirections, in which the client or the authorization server directs the resource owner's user-agent to another destination_ (https://www.rfc-editor.org/rfc/rfc6749#section-1.7) and regarding authorization code grant it says _this is a redirection-based flow_ (https://www.rfc-editor.org/rfc/rfc6749#section-4.1)
So, if one MNO makes a redirect to an endpoint where HE is working its an option, if another sees public IP is a different option. In both options any further redirects may occur (as it usually happen)
- Also notice that Network Auth is only applied to Authorization Code flow, so I don't think what you say about request/response applies, it will always be based on redirects as it said by standard
- Finally, notice in no place its said that this is a universal and always valid solution. This is one of the solutions availabie that may apply or not to the APIs depending on conditions. To go to the obvious case, it won't work if there is no Mobile Network connection, Network-based auth will not be possible. 
Regarding 2 we are alligned, CAMARA won't specify specific mechanisms for Network Based Auth, that is not relevant for developer

shilpa-padgaonkar:
@DT-DawidWroblewski : The statements you have referred above are **NOT** the technical ruleset.

The technical ruleset in the [CAMARA-API-access-and-user-consent.md](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#user-identity) documentation only points out which kind (auth-code or ciba) of 3-legged flow to use once a 3-legged flow is confirmed.

To be precise, **the technical ruleset** includes only the following 2 lines:

- Technical ruleset for the Frontend flow

If all API usecases point to the need of On-net scneario and where the consumption device and authentication device are the same, the front end flow must be used. eg. NumberVerification

- Technical ruleset for the Backend flow

If some usecase/s for an API point to off-net scenarios and where consumption and authenticaiton devices could be different, the backend flow must be used.

diegogonmar:
@DT-DawidWroblewski  the points you mention are not part of the file updated in this PR.
I think we should not discuss parts of the document not changed by this pull request, nor other documents. 

Anyway, AuthCode grant mandates to support redirects, its not something we're specifying here. I quoted that before:
OAuth RFC says: This specification makes extensive use of HTTP redirections, in which the client or the authorization server directs the resource owner's user-agent to another destination (https://www.rfc-editor.org/rfc/rfc6749#section-1.7) and regarding authorization code grant it says this is a redirection-based flow (https://www.rfc-editor.org/rfc/rfc6749#section-4.1)
Also HE may work with or without redirects, but in any case does not preclude the AuthCode flow as a whole to involve (further) redirections.

I don't fully understand last point. There are other ways of doing it, but here we're documenting the network based authentication mechanism.

jpengar:
As we talk in the last working group call conference (20/09), the original goal of this issue and PR was to include some information about network based authentication in the `CAMARA-AuthN-AuthZ-Concept.md` document. This type of authentication is specifically mentioned in the Number Verify API and it was decided to include the related information in this WG for any CAMARA participant to use as a reference. Network-based authentication is a requirement for the operator, and the specific mechanism used (based on IP, header encryption, etc.) is up to the operator. This is not something to standardize here.
 
The current changes provided by PR #55 should already cover this issue. However, @DT-DawidWroblewski points to his comments in the PR regarding the topic that with HE (Header Enrichment) HTTP is involved and therefore may not work, at least the developer needs to be aware of that. That's a fair point to be considered. 

@DT-DawidWroblewski PR could be revisited for further clarification on this issue. Would you please contribute to this PR with any clarification you feel is needed in `CAMARA-AuthN-AuthZ-Concept.md` document.?

jpengar:
As per yesterday's workgroup call (04/10), I propose to merge this PR as it covers the original scope of the issues and the PR has been open for weeks. If Dawid (DT) or any other CAMARA participant wants to clarify something else on "header enrichment" or any other topic, a new issue/PR can be opened and they can contribute on their own schedule. There is no problem at all, but I prefer to merge the PR and not condition the closure of issue #45 just for further clarifications that could be added later.
@rartych could you have a look at it and give your approval so we can merge the PR?

rartych (approved):
LGTM

PR #53, opened 2023-07-27
IdentityAndConsentManagement meeting notes 2023-07-26

jpengar:
#### What type of PR is this?
* documentation

#### What this PR does / why we need it:

MoM of the call held in 2023-07-26.

#### Which issue(s) this PR fixes:

None.

PR #62, opened 2023-09-14
Create CAMARA-API-AuthN-Inventory.md

### LINKED ISSUE #57: Align `securitySchemes` and `security` of CAMARA API specs with IdentityAndConsentManagement WG
This issue is intended to define a single common/standard OpenID Connect security scheme for CAMARA API specs aligned with the OIDC FE/BE flows defined in https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/user-consent.md#authorization-flows--grant-types as the access flows to CAMARA APIs.

e.g.
```
components:
  securitySchemes:
    openId:
      type: openIdConnect
      openIdConnectUrl: https://example.com/.well-known/openid-configuration
```

This ensures that the **CAMARA API specifications are consistent with each other and with the IdentityAndConsentManagement work**: 

- Any reference to other `securitySchemes`, such as client credentials, would generally be removed. Client credentials access (2-legged) must only be used _"when user resource is NOT involved & the client belongs to confidential category"_.
- Those APIs without defined scopes should define the corresponding scopes that protect the API operations, and they must be associated with the OpenID Connect security scheme by means of the `security` property.

**This has already been proposed in Commonalities in: https://github.com/camaraproject/Commonalities/issues/53 where the proposal is very detailed. So we can just refer this issue in IdentityAndConsentManagement to issue #53 in Commonalities. The same Commonalities issue is not only intended to unify the usage of `securitySchemes` and `security` fields, but also to define a common scope naming format for CAMARA API specifications.**

shilpa-padgaonkar:
Partly Fixes camaraproject/IdentityAndConsentManagement#57 

#### What type of PR is this?

* documentation

This document will provide an overview of security schemes for Camara APIs

<!-- Automatically closes linked issue when PR is merged.
Partly Fixes camaraproject/IdentityAndConsentManagement#57 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/CAMARA-API-AuthN-Inventory.md
+++ b/documentation/CAMARA-API-AuthN-Inventory.md
@@ -1,14 +1,14 @@
 ## CAMARA-API-AuthN-Inventory.md
 
-| API Family | API Name | Scopes (not needed for oauth2-cc) | Security-scheme name | Type | Flow |
+| API Family | API Name | Scopes (not needed for oauth2-cc) | Security-scheme name | Type | Recommended Flow |
 | -------- | -------- | ------ | ------------- | ---- | ---- |
-|| qod| qod-sessions-read, qod-sessions-write, qod-sessions-delete, qod-profiles-read |oauth2-cc | oauth2 | clientCredentials |
-|| number-verification |number-verification-verify-read, number-verification-share-read | openId-ac | openIdConnect | authorizationCode |
-|| device-status |xxxx (To be added)| openId-cb | openIdConnect | ciba |
-| DeviceLocation | location-verification |xxxx (To be added)| openId-cb | openIdConnect | ciba |
-| DeviceLocation | location-retrieval |xxxx (To be added)| openId-cb | openIdConnect | ciba |
-|| device-identifier ||oauth2-cc | oauth2 | clientCredentials |
-| OTPValidation | one-time-password-sms||oauth2-cc | oauth2 | clientCredentials |
-|  CarrierBillingCheckOut| carrier-billing-payments| carrier-billing-payments-create, carrier-billing-payments-prepare, carrier-billing-payments-confirm, carrier-billing-payments-cancel, carrier-billing-payments-read | openId-ac | openIdConnect | authorizationCode |
+|| qod| qod-sessions-read, qod-sessions-write, qod-sessions-delete, qod-profiles-read |oauth2 | oauth2 | clientCredentials |
+|| number-verification |number-verification-verify-read, number-verification-share-read | openId | openIdConnect | authorizationCode |
+|| device-status |xxxx (To be added)| openId | openIdConnect | ciba |
+| DeviceLocation | location-verification |xxxx (To be added)| openId | openIdConnect | ciba |
+| DeviceLocation | location-retrieval |xxxx (To be added)| openId | openIdConnect | ciba |
+|| device-identifier ||oauth2 | oauth2 | clientCredentials |
+| OTPValidation | one-time-password-sms||oauth2 | oauth2 | clientCredentials |
+|  CarrierBillingCheckOut| carrier-billing-payments| carrier-billing-payments-create, carrier-billing-payments-prepare, carrier-billing-payments-confirm, carrier-billing-payments-cancel, carrier-billing-payments-read | openId | openIdConnect | authorizationCode |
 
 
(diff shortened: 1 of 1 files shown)


shilpa-padgaonkar:
@hdamker @DT-DawidWroblewski @eric-murray : I have added as part of this inventory, examples that include APIs where you are code-owners (I couldn't tag all code owners of these subprojects as not all are subscribed to this subproject). Kindly verify if the security schemes specified here are fine after alignment within your subprojects.

jlurien:
In QoD we currently have both clientCredentials and authorizationCode. We have to align everything everywhere and decide which criteria must be applied. First check is to review whether the operation deals with personal information in order to decide when 3-legged flow applies.

This topic is also being discussed in Commonalities: [issue #53](https://github.com/camaraproject/Commonalities/issues/53), and [PR #57](https://github.com/camaraproject/Commonalities/pull/57)

shilpa-padgaonkar:
@jlurien Fully agree. Hence, this inventory. The naming can be finalized as you said in https://github.com/camaraproject/Commonalities/pull/57 and inventory can be maintained here using https://github.com/camaraproject/IdentityAndConsentManagement/pull/62/files as it makes more sense to keep it in this subproject. WDYT? We can also wait for more people to comment and also for @jpengar to be back from his vacation.

jlurien:
Yes, I think we are totally aligned on this. It is OK to centralize all I&C stuff here. In the Commonalities guidelines there is also a chapter about security. We may decide to ref that part to some doc here, to avoid duplication if necessary.

DT-DawidWroblewski:
Hi @shilpa-padgaonkar !

for SimSwap: OAUTH2.0 Client Credentials & OpenID Connect CIBA
for NumberVerification: OpenID Connect
for OTPValidation: OAUTH2.0 Client Credentials

shilpa-padgaonkar:
@DT-DawidWroblewski : For simswap I see a different comment (https://github.com/camaraproject/IdentityAndConsentManagement/issues/63#issuecomment-1720731103)  from you as compared to the above suggestion. 

The idea of this inventory is to align on a single flow for api-scope/s. I have extended the list with the other 2 APIs you have provided details for. 

After the flow has been aligned for sim swap in the subproject, feel free to make a commit directly to the PR.

diegogonmar:
Hi @shilpa-padgaonkar, the idea is to have single one? I could think of APIs that may work both in CIBA and AuthCode, as both are 3-legged and use cases may exist enabling valid for both flows.

shilpa-padgaonkar:
@diegogonmar Yes the idea is to have a single flow. So if there are cases where both CIBA or AuthCode could be used as you have hinted above, the technical ruleset could be to determine which is one to add.

Technical ruleset for the Frontend flow

If all API usecases point to the need of On-net scneario and where the consumption device and authentication device are the same, the front end flow must be used. eg. NumberVerification

Technical ruleset for the Backend flow

If some usecase/s for an API point to off-net scenarios and where consumption and authentication devices could be different, the backend flow must be used.

diegogonmar:
Yes, you're right, that was the agreed wording. I'd rather open different issue in case I think further discussion is needed about the ruleset, to cover scenarios where maybe 2 grants apply

chrishowell:
Is there a reason why Carrier Billing requires the OIDC AuthCode flow? I can imagine a scenario where the consumption device and authorisation device are different, and the Technical Ruleset would imply that the Backend Flow should be used?

shilpa-padgaonkar:
I have changed this PR to draft, so that we wait until subprojects are ready to piggyback their cleanup results back here & to ensure that we do not start any subproject related questions as a result of this PR review here.

ECORMAC:
In Number Verification we have adhered to the use of OIDC security scheme as per commonalities guidelines (based on work in this group I beleive). However OIDC as per standard would also entail the use / delivery of identity tokens. In our case (Number Verification) we don't see the need for identity tokens (but see the need for Authorization code grant). I checked with commonalaties if there was some recommendation / principle that we could use to both adhere to OIDC but not use / implement identity tokens., However they pointed at this group for guidance.

jpengar:
@ECORMAC Thank you for your comment. But I think you should open a new issue in https://github.com/camaraproject/IdentityAndConsentManagement/issues as this PR is currently on hold as a draft and is not directly related to the topic you raised above.

Issue #32, opened 2023-05-23
The concept of purpose


jpengar:
**Problem description**

CC @sebdewet 

Hi,

regarding the concept of purpose, we (Orange) would like to correctly understand the benefits to add the purposes to the OIDC standard. We already have a group component using the standard and there will be impacts in this case.
Our understanding is that if purpose is studied/proposed, it means that the standard with scopes isn't sufficient for all the use case needed in the camara perimeter.
However, we currently don't know which use case won't be covered by the standard. Or is it just in order to "provides a human-friendly description" ?
We also wonder if adding the purpose would complicate the standard with another layer to define, administer, manage changes, agree on vocabulary etc.

thank you in advance for the clarifications

**Additional context**
This is a technical challenge open to discussion.

sebdewet:
Hi,

> as users may give consent to access those API resources for a functional motive, but not for a commercial/advertising motive, for example

the fact that the End-User will be given the option to have the OpenID Provider decline to provide some or all information requested by RPs is already in the standard, to share only phone and email and not address for example. The benefits would be to give consent to share email for commercial motive, and email and phone for another motive ? What would be the use cases associated to this functionnality ? 

> We are open to defining a new 'purpose' query parameter in the authorization request, as we originally suggested in a discussion held in Commonalities a few months ago, or we could agree on a technical solution to provide the name of the purpose in the existing standard 'scope' parameter (e.g. scope="purpose:xyz").

Could we already define the impact of each scenario ? pros and cons ?

> Using just OIDC scopes in the Auth request and defining N scopes for the same API/resource/operation would require to change API definition to add new scopes every time a new usage (with a new purpose) is defined for the applications.

a new purpose will still be dependant of the existing scopes, thus I don't see the benefits.

> You could easily control for example if the same API can be accessed or not depending on the usage the a same application want to do of that API by means of the purpose indicated by the application when requesting the access token.

same behaviour withs scopes ?

> And purposes can map multiples scopes corresponding also to multiple APIs, so you could actually group all accessible resources under the same purpose definition.

I think i like this benefit, but could you detailed it ?

is the purpose is finally an equivalent of several endpoint as userinfo for example, which is a also multiple scope values ?

jpengar:
[...]

PR #64, opened 2023-09-20
Telefonica Orange Consent&Purpose comparison table

bigludo7:
In order to move forward on issue https://github.com/camaraproject/IdentityAndConsentManagement/issues/32 please see a table comparing both proposal via faq.

**Not to be merge**
**Must be completed by Telefonica team**

New question on the faq welcome.

#### What type of PR is this?

* documentation

Move forward on isue https://github.com/camaraproject/IdentityAndConsentManagement/issues/32 

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```


bigludo7:
Would like to add @diegogonmar as reviewer but not able to add reviewer.

documentation/SupportingDocuments/Purpose Consent Proposal comparison.md
```diff
@@ -0,0 +1,34 @@
...
+| Do link between OPGW product and purpose is visible in Telco Catalog? |  |  In the catalog, it could be possible to browse OPGW product per category. Several categories could be created, as such 'purpose' could be defined as a category, and purpose  value could be defined as category value (in order to get for example OPGW Product related to Fraud prevention & detection) but this is only informative information.  |
+| What are country market required agreement before to launch?|   | Telco must agree on the OPGW product to be launched in the country as a minimum 'critical mass' must be available. Additionally Telco must indicate for each API scope the legal base associated with purposes for this country.|
+| What is expected during Application onboarding?|   | When a developer wishes to uses CAMARA API, developer first onboards an application. Mandatory information to onboard an application is to provide the purpose (or the list of purposes) associated with this application. |
```
shilpa-padgaonkar, line 19: In addition to the purpose, doesn't the developer need to choose which API products (with which exact scope/s) he intends to use within the app?

documentation/SupportingDocuments/Purpose Consent Proposal comparison.md
```diff
@@ -0,0 +1,34 @@
...
+| How the legal consent displayed to the end user is defined and provided? | | Telco  associates a legal wording for each scope design-time. Then on the fly the legal notice is generated and displayed to the front end. <br> <br > As an example, simswap-verify-read technical scope is associated to wording:  'Verify latest SIM swap Date', number-verification-verify-read to 'Verify your MSIDN from network'. The request `POST /authorize` with `scope=FraudDetectionAndPrevention:simswap-verify-read` triggers following notice: <br> _"For fraud detection and prevention, application NiceApp request your consent to perform following action(s): <br> - Verify latest SIM swap Date <br> - Verify your MSIDN from network"_
```
shilpa-padgaonkar, line 21: With this solution, would it be also allowed to send within the same post request, several different purposes? For eg. FraudDetectionAndPrevention and ServiceOptimization?

bigludo7: @shilpa-padgaonkar No but I'm not sure about the UC you have in mind. When an `authorize` is send for claiming data access to end user data from an app I'm not sure how we can have several purpose. I understood we can have an app with several purpose but for one given call only one purpose should be passed. I update the table accordingly.

bigludo7 (commented):
Adding one question and one precision following @shilpa-padgaonkar comments

documentation/SupportingDocuments/Purpose Consent Proposal comparison.md
shilpa-padgaonkar: I am glad to hear this! We can ONLY support the scenario if exactly one purpose is sent for a given call.

documentation/SupportingDocuments/Purpose Consent Proposal comparison.md
```diff
@@ -0,0 +1,35 @@
...
+| What are country market required agreement before to launch?|   | Telco must agree on the OPGW product to be launched in the country as a minimum 'critical mass' must be available. Additionally Telco must indicate for each API scope the legal base associated with purposes for this country.|
+| What is expected during Application onboarding?|   | When a developer wishes to uses CAMARA API, developer first onboards an application. Mandatory information to onboard an application is to provide the purpose (or the list of purposes) associated with this application. |
+| What is expected during API onboarding?|   | When a developer wishes to use a CAMARA API (prequisite developer has onboarded an app - previous point), it has to contractually agrees with the Hyperscaler/agregator/telco about Open Gateway product purchase. There is not tight coupling with purpose at this stage. Of course in the catalog the Telco could recommend some API for some purpose but API (and associated Open Gateway product) and purpose are not associated.  |
```
shilpa-padgaonkar, line 20: @bigludo7 Even if product is not tightly coupled with purpose, wouldn't the contractual agreement need to eventually tie together the App+Purpose+API-Product(with exact scopes)? Wouldn't application onboarding be a good place to bring these things & contractually agree on them?

bigludo7: @shilpa-padgaonkar This is a good point to discuss. 
For me it is important to have a tight coupling between app and purpose. I guess as telco we have to make our 'homework' when a developer request to onboard an app for a given purpose.

Then when we sell API, we sell them for a given app (or a group of app) of one given developer. You have there app <-> purpose(s) and app <-> APIs. I do not have direct link between API and purpose but I'm able to know that I sold these APIs for this app which as this purpose.
Then in the `authorize` request we mandate the purpose (and I got the app indirectly via the credential)  - notably to check that we expect this purpose usage from this app (I do want to have request for a purpose not covered in the agreement). 
hope it makes sense.

shilpa-padgaonkar: @bigludo7 
Yes exactly.  We should be able to derive that the API is used for a specific app which has agreed to a specific purpose.

What we have in mind is that at the time of app onboarding, the developer should select the purpose. He also selects the relevant API-product for the app and selects the list of relevant scopes. It is important for us that we are able to derive the legal basis at the time of onboarding the app with this combination (purpose-scopes). This combination of purpose and scopes has to lead to a single "legal basis". So when the authorize request is sent, we can check that the sent combination of purpose and scopes was exactly as was agreed during this onboarding or not. If it is not the same (if for eg. the request contains less or more scopes or if the purpose is different) we will reject the said request.

bigludo7: @shilpa-padgaonkar 
Oups my bad.... should be 

> "I do **not** want to have a request for a purpose not covered in the agreement"

For us the purpose is part of the agreement for an application. This is a critical information.
Then the developer onboard for API for this application. As the on-boarded API are for **this** application they inherit for the application purpose.
As we have then the purpose (from the app), the technical scope (from the api yaml) and the country(ies) we can inform the developer about the legal base that will be applied. This legal base will be enforced in our Authorization server.

In my catalog I do not have a tight coupling between API product and purpose.
For a API usage (from an application) I have tight coupling with the application purpose and mechanistically to the legal base to apply.

documentation/SupportingDocuments/Purpose Consent Proposal comparison.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Proposal presentation
+
+| Topic     | Telefonica proposal | Orange proposal |
+| --------- | ------------------- | --------------- |
+| What is the Open Gateway (OPGW) Product granularity? |  |  Open Gateway product matches 1-1 with a CAMARA API. For example, as of now, the CAMARA Device location API family features 2 APIs (2 yaml): `location-retrieval` and `location_verification` : we consider them as 2 separate OPGW product. The OPGW product covers completely the API (all the resources of the API). This is an important point because it means that API granularity must not be considered only from a technical aspect but also from a business aspect (which is already the case for CAMARA API). <br> The OPGW product is **not** strong-coupled with a purpose but independent of it. However in the catalog it is possible to recommend OPGW product with purpose (see next point).|
```
jpengar, line 16: Telefónica disagrees. For example, the SIM Swap API has two different endpoints (/check and /retrieve-date) and these endpoints will be different products as they require different legal basis. There will not be one SIM Swap product for the whole API (all resources of the API).

shilpa-padgaonkar:
@bigludo7 Thanks for the quick responses to my review comments.  There are a couple of points where I might have a slight variation to add/specify. I will add a third column here and at least add points for a row or 2 where I see these small variations.

jpengar:
@bigludo7 @shilpa-padgaonkar We've updated the table with Telefonica feedback. :)

We are a bit concerned because many of the topics in the table above are actually not technical topics, but business topics related to product and application definition/onboarding. Here we should only try to agree on the technical approach to declare the purpose when an application requests access to CAMARA APIs (/authorise request) and how operators should handle the authorisation request. And this should not be tied to the OpenGateway product definition, because this may change, or because CAMARA APIs may be used outside the OpenGateway context.

bigludo7:
@jpengar Thanks for the contribution.
I did a mistake with merging the PR with @shilpa-padgaonkar  - I'm aware of this and will fix this today.

bigludo7 (commented):
Added Telefonica proposal

rartych:
@bigludo7 I propose to add numbering in Topics column - this can help to reference topics in the discussion. WDYT?

bigludo7 (commented):
Merged @shilpa-padgaonkar contribution

bigludo7:
> @bigludo7 I propose to add numbering in Topics column - this can help to reference topics in the discussion. WDYT?

@rartych Yes sure !

bigludo7 (commented):
Add line identifier + Orange answer for 11 & 12.

jpengar:
@bigludo7 @shilpa-padgaonkar I think we can close this PR already as the agreement reached is already described and being discussed in #32

bigludo7:
@jpengar Works for me.

shilpa-padgaonkar:
@jpengar Fine for me.

PR #66, opened 2023-09-22
IdentityAndConsentManagement meeting notes 2023-09-20

jpengar:
#### What type of PR is this?
* documentation

MoM of the call held in 2023-09-20.

None.

### DIFF
--- a/documentation/MeetingMinutes/MOM-2023-09-20.md
+++ b/documentation/MeetingMinutes/MOM-2023-09-20.md
@@ -1,0 +1,123 @@
+# Identity and Consent Management meeting
+
+#### *20th of September 2023*
+
+<br>
+
+## Attendees
+
+
+| Companies | Attendees |
+| --------- | --------- |
+| Telefónica | Jesús Peña, Diego Gonzalez |
+| GSMA      | Daniel Okonkwo, Mark Cornall |
+| Vodafone  | Eric Murray, Sönke Peters, Klaus Reifenrath |
+| Singtel   | Foo Ming Hui |
+| T-Mobile US | Murat Karabulut |
+| Rohde & Schwarz | Tim Kittel |
+| Deutsche Telekom AG | Shilpa Padgaonkar, Dawid Wróblewski, Rafal Artych |
+| Gapask  | Rajesh Murthy |
+| Orange    | Robert Ludovic |
+| Microsoft | Landon Cox |
+
+<br>
+
+## Agenda
+
+1. Recent updates & recap
+2. Issues and PRs. Priority discussions:
+    - Discussion on Issue #32. "The concept of purpose".
+    - Closure of Issue #45. "Add missing details in authN-authZ doc".
+    - Closure of Issue #63. "Usage of OAUTH2.0 client credentials in SimSwap API".
+    - Discussion on Issue #57 "Align securitySchemes and security of CAMARA API specs with IdentityAndConsentManagement WG" and PR #62 "Create CAMARA-API-AuthN-Inventory.md".
+3. Action Points 
+4. AoB
+
+<br>
+
+## Recent updates & recap
+
+Telefonica presents the status of WG:
+
+- Issues closed since last call 06/09
+  - N/A - No issues closed in the last two weeks.
+
+- PRs merged since last call 06/09
+  - PR #60 "IdentityAndConsentManagement meeting notes 2023-09-06"
+  - PR #61 "typos and reference to CIBA"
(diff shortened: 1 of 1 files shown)


jpengar:
@DT-DawidWroblewski FYI. I couldn't set you as a reviewer.

rartych (approved):
LGTM

documentation/MeetingMinutes/MOM-2023-09-20.md
```diff
@@ -0,0 +1,123 @@
...
+| [#32](https://github.com/camaraproject/IdentityAndConsentManagement/issues/32) ([PR#64](https://github.com/camaraproject/IdentityAndConsentManagement/pull/64)) | Orange | **The concept of purpose**<br> There is still no agreement on the technical solution to declare a "purpose" when accessing the CAMARA API. This issue is already becoming a development bottleneck and needs to be resolved sooner rather than later. Telefonica proposes to set a deadline to try to find an agreement or an intermediate solution between the proposal of Telefonica and Orange. Hopefully we manage to do so, this is the desired way. But if there is no final agreement, it would be necessary to vote it according to CAMARA governance rules.<br>New PR #64 opened by Robert Ludovic (Orange) including a table to compare both solutions. Telefonica will complete it by Friday (22/09), then DT will review and add comments, maybe add new columns.<br>**Deadline will be the end of next week: 29/09. If no agreement is reached, the voting process would be organized for the week of 02/10.** |
```
shilpa-padgaonkar, line 102: @jpengar What is meant by voting process here? Is this discussion + voting in TSC or simply open voting?

jpengar: @shilpa-padgaonkar Whatever is in line with CAMARA governance and TSC responsibilities. We can agree on the specific process to vote this properly. Whether it goes through TSC or not, I do not have a strong opinion on that. I don't know all the details of CAMARA governance, it's something that would have to be agreed if necessary. Hopefully we can come to an agreement next week to avoid a vote.

shilpa-padgaonkar: @jpengar Thanks. Since commonalities and id-consent are not typical subprojects and as decisions made in these 2subprojects/wg impact all other subprojects, I just wanted to make sure that we do not intend to conduct any voting which is restricted only to "id-consent" members.  Hopefully this topic is resolved like you said by next week and we dont need to vote. If not, let's take it to TSC and trigger a discussion there to check if vote needs to be an open vote or if TSC can get it resolved without a vote.

documentation/MeetingMinutes/MOM-2023-09-20.md
```diff
@@ -0,0 +1,123 @@
...
+| [#63](https://github.com/camaraproject/IdentityAndConsentManagement/issues/63) | Deutsche Telekom AG | **Usage of OAUTH2.0 client credentials in SimSwap API**<br> Moved to SIM Swap project as requested by Shilpa (DT). However, it was clarified that this issue actually stems from a discussion in the SIM Swap subproject where it was requested to move the discussion to this WG in order to clarify the agreed guidelines regarding the use of client credentials in CAMARA API access. The issue will be moved back to this WG and it is agreed to close it, as it is already clarified in the issue comments that client credentials do not apply to the SIM Swap API, nor to any other API processing personal user data. Dawid (DT) pointed out that in his opinion the introduction of CIBA into SimSwap looks like a "downgrade" of existing solutions that customers are using. But CIBA flow fits perfectly for the mass market,so it should be fine for CAMARA.<br>Issue will be moved back to this WG and CLOSED according to what is mentioned above.  |
```
DT-DawidWroblewski, line 99: I'd appreciate having a clear statement from legal/regulatory teams about handling user-related data in Camara APIs, including SimSwap API. Today, it is not clear how a legitimate interest use case has to be handled - should we use opt-out for every scenario or not, e.g. bank that delivers banking services and a bank's legitimate interest is to use a sim swap API for fraud prevention, so bank customer cannot opt-out from this feature.

jpengar:
Hi, I will merge the notes today if there are no further objections to what is specific to the meeting notes.

bigludo7 (approved):
Thanks.
I understood the point about moving forward quickly but I'm still puzzled about the voting there are we do not have an explicit question to be submitted;
With this comment in mind, I'm ok for the meeting minutes.

jpengar:
@bigludo7 That's OK. This PR is just to provide some notes on what was discussed and mentioned in the last working group call (for the record). We didn't agree or discuss details of how this vote would work during that call, so it would be something we would need to review and agree if needed. As I also commented to @shilpa-padgaonkar above where she suggested this to be raised at the TSC. Hopefully we can resolve this before we need to do all this.

PR #109, opened 2024-01-16
Add rule for requesting scopes for client credentials

### LINKED ISSUE #92: Scope definition for Client Credentials
**Problem description**
I raised this as a [discussion](https://github.com/camaraproject/IdentityAndConsentManagement/discussions/81), but got no responses. So I am raising it as an issue.

The current [CAMARA APIs Access and User Consent Management Guidelines](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#purposes) say **how** to declare a purpose within a scope definition:

> ### Applying purpose concept in the authorization request
> In order to declare a purpose when accessing the CAMARA APIs, it is strongly recommended that the 'scope' parameter is set to:
> `dpv:<dpvValue>#<technicalParameter>`

but are not clear about **when** a purpose needs to be declared. In particular, there is no explicit definition of how scopes are constructed for authorisation using client credentials. It may seem obvious that the scope in this case should be the `technicalParameter` alone, but that does not appear to be defined anywhere.

**Expected action**
It should be clarified that a purpose only needs to be defined when using the one of the OpenId Connect authentication schemes (`grant_type` equals `authorization_code` or `urn:openid:params:grant-type:ciba`) and not when `grant_type` equals `client_credentials`.

**Additional context**

eric-murray:
#### What type of PR is this?
* documentation

Defines how scope requests are constructed when using the client_credentials grant type

Fixes #92 

None

#### Changelog input
```
 release-note
 - Add rule for requesting scopes for client credentials
```

#### Additional documentation 
None

### DIFF
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -80,7 +80,7 @@ A purpose must be declared within the requested scope for the following grant ty
 - authorization_code
 - urn:openid:params:grant-type:ciba
 
-For the client_credentials grant type, the requested scope must be set directly to the value defined for the relevant endpoint within the OAS ("YAML") specification. "Wild card" scopes (i.e. specifying only the API name) are not valid for this grant types.
+No purpose is required for the client_credentials grant type, as this grant type can only be used when no personal user data is processed. The requested scope must be set directly to the value defined for the relevant endpoint within the OAS ("YAML") specification. "Wild card" scopes (i.e. specifying only the API name) are not valid for this grant type.
 
 In order to declare a purpose when accessing the CAMARA APIs, it is strongly recommended that the `scope' parameter is set to:
 
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -76,6 +76,12 @@ A purpose declares what the application intends to do with a set of personal inf
 
 ### Applying purpose concept in the authorization request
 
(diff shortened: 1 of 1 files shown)


documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -76,6 +76,12 @@ A purpose declares what the application intends to do with a set of personal inf
 
 ### Applying purpose concept in the authorization request
 
+A purpose must be declared within the requested scope for the following grant types:
+- authorization_code
+- urn:openid:params:grant-type:ciba
+
+For the client_credentials grant type, the requested scope must be set directly to the value defined for the relevant endpoint within the OAS ("YAML") specification. "Wild card" scopes (i.e. specifying only the API name) are not valid for this grant types.
```
jpengar, line 83:
```suggestion
No purpose is required for the client_credentials grant type, as this grant type can only be used when no personal user data is processed. The requested scope must be set directly to the value defined for the relevant endpoint within the OAS ("YAML") specification. "Wild card" scopes (i.e. specifying only the API name) are not valid for this grant type.
```

jpengar (commented):
I'm pretty much ok with PR content. I just added a suggested change with an additional clarification.

jpengar (approved):
After https://github.com/camaraproject/IdentityAndConsentManagement/pull/109/commits/7675e3e027b594cdf16bbced5374a131f0cc840e. LGTM

jpengar:
Merged as agreed in yesterday's WG meeting call (17/01/2024)

PR #113, opened 2024-01-25
Add CAMARA OIDC Profile

### LINKED ISSUE #78: clarify the "credentials" notion in CIBA workflow authentication
**Problem description**
the value "credentials" in the [CIBA workflow authentication](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#ciba-flow-backend-flow) isn't clear

**Expected action**
Define clearly what is expected in the notion "credentials" (basic auth, client_id, token etc.)

**Additional context**
in the [openid standard documentation](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html), credentials are defined with the notion of client_notification_token.
In the Orange implementation, only client_id is used.

### LINKED ISSUE #90: OIDC guidance (what needs to be adhered to?)
**Problem description**
In Number Verification we have adhered to the use of OIDC security scheme as per commonalities guidelines. 
However OIDC as per standard would also entail the use / delivery of identity tokens. In our case (Number Verification) we don't see the need for identity tokens (but see the need for Authorization code grant). 
Having checked with commonalities for some recommendation / principle that could be used to both adhere to OIDC but not use / implement identity tokens, they indicated that IdentityAndConsentManagement should be able to provide some guidance (that Number Verification and commonalities) can use.  

**Expected action**
Clarification on the directive to adhere to OIDC - what parts are mandatory, if all parts are not mandatory is for example a compliance statement necessary in each workgroup?

**Additional context**

### LINKED ISSUE #104: Alignment of Client Credential authentication / authorization using API Gateway Pattern
**Problem description**
Referring to the following:
[CIBA Backend Flow](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#ciba-flow-backend-flow)

```
POST /bc-authorize HTTP/1.1
Authorization: Basic {Credentials}
Content-Type: application/x-www-form-urlencoded

login_hint=tel%3A%2B346xxyyyzzz&
… 
```

The authorization should follow [API Gateway Pattern](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-AuthN-AuthZ-Concept.md#api-gateway-pattern-) with [Client Credential Grant](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-AuthN-AuthZ-Concept.md#client-credential-grant) access token usage, instead of Basic authentication.

**Expected behavior**
Align with API Gateway Pattern mentioned above.

**Alternative solution**
N.A.

**Additional context**
GSMA has a concept of a north star where the API consumer authentication / authorization is the same implementation for all Operator Open Gateway CIBA implementation and only a change in credentials is sufficient for a change in API provider.

garciasolero:
#### What type of PR is this?

* documentation

Add first draft of CAMARA OIDC Profile:

- Authorization flows:
  - Authorization Code Flow.
  - Client Initiated Back Channel Authentication (CIBA).
  - Client Credentials.
- Current definition of scope parameter format applicable to CAMARA APIs that could be evolve in the future. 
- ID Token content proposal based on OIDC standard.  
- Client Authentication.

- Fixes #78
- Fixes #90 
- Fixes #104 

Pending topics to be discussed for inclusion in future versions of the profile: 

- Purpose/Scope redefinition
- PKCE
- `request` parameter in CIBA flow
- DPoP

#### Changelog input

N/A

#### Additional documentation 

N/A

### DIFF
--- a/documentation/CAMARA-OIDC-profile.md
+++ b/documentation/CAMARA-OIDC-profile.md
@@ -336,7 +336,7 @@ Content-Type: application/json
 ```
 #### Error Response
 
-The authorization server responds with an HTTP 400 (Bad Request) status code (unless specified otherwise) and includes the following parameters with the response:
+The authorization server responds with an HTTP 400 (Bad Request) status code (unless specified otherwise) and includes the following parameters within the response body:
 
 * **error**
 
--- a/documentation/CAMARA-OIDC-profile.md
+++ b/documentation/CAMARA-OIDC-profile.md
@@ -40,7 +40,7 @@
 
 ## Introduction
 
-Telco network capabilities exposed through APIs provide great value to customers. By simplifying the complexity of telco networks with APIs and making the APIs available across telco networks and countries, CAMARA enables easy and seamless access. To ensure seamless interoperability and increased security, this technical document introduces the OpenID Connect (OIDC) profile, which is specifically tailored for CAMARA APIs.
+Telco network capabilities exposed through APIs provide great value to customers. By simplifying the complexity of telco networks with APIs and making the APIs available across telco networks and countries, CAMARA enables easy and seamless access. To ensure seamless interoperability and increased security, this technical specification introduces the OpenID Connect (OIDC) profile, based on several standards, which is specifically tailored for CAMARA APIs.
 
 OpenID Connect is a widely adopted standard for user authentication and authorization in the context of APIs. This document builds on the basic principles of OIDC and outlines CAMARA-specific guidelines to address CAMARA APIs scenarios and use cases.
 
@@ -100,7 +100,7 @@ The CAMARA profile uses the following OpenID Connect request parameters with the
 
 * **redirect_uri**
 
-  REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.1). When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
+  REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.1). When using this flow, the Redirection URI MUST use the https scheme; however, it MAY use the http scheme only for non-production environments. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
 
 
 * **state**
@@ -348,7 +348,7 @@ The authorization server responds with an HTTP 400 (Bad Request) status code (un
 
     * **_invalid_client_**
 
-      Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The authorization server MAY return an HTTP 401 (Unauthorized) status code to indicate which HTTP authentication schemes are supported. If the client attempted to authenticate via the `Authorization` request header field, the authorization server MUST respond with an HTTP 401 (Unauthorized) status code and include the `WWW-Authenticate` response header field matching the authentication scheme used by the client.
+      Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The authorization server MAY return an HTTP 401 (Unauthorized) status code to indicate which HTTP authentication schemes are supported.
 
     * **_invalid_grant_**
 
@@ -651,7 +651,7 @@ The authorization server responds with an HTTP 400 (Bad Request) status code (un
 
     * **_invalid_client_**
 
-      Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The authorization server MAY return an HTTP 401 (Unauthorized) status code to indicate which HTTP authentication schemes are supported. If the client attempted to authenticate via the `Authorization` request header field, the authorization server MUST respond with an HTTP 401 (Unauthorized) status code and include the `WWW-Authenticate` response header field matching the authentication scheme used by the client.
+      Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The authorization server MAY return an HTTP 401 (Unauthorized) status code to indicate which HTTP authentication schemes are supported. 
 
     * **_invalid_grant_**
 
@@ -718,7 +718,7 @@ The client makes a request to the token endpoint by adding the following paramet
 
 * **scope**
 
-  OPTIONAL. A space delimited and a case-sensitive list of ASCII strings.
+  REQUIRED. A space delimited and a case-sensitive list of ASCII strings. If the scope parameter is not present, the behavior is entirely unspecified (i.e., the authorization server may return an `invalid_request` error or accept the request by assigning it a default value).
 
 
 The client MUST authenticate with the authorization server as described in [Client Authentication Section](#client-authentication).
@@ -841,26 +841,7 @@ The Authorization Request MUST contain the `scope` parameter, which is a space d
 
 For OpenID requests, the scope values MUST include `openid`, to indicate that the request is an OpenID Connect request, followed by other values depending on the specific CAMARA services being requested.
 
-A purpose must be declared within the requested scope for the following grant types:
-
-* _authorization_code_
-
-* _urn:openid:params:grant-type:ciba_
-
-No purpose is required for the `client_credentials` grant type, as this grant type can only be used when no personal user data is processed. The requested scope must be set directly to the value defined for the relevant endpoint within the OAS ("YAML") specification. "Wild card" scopes (i.e. specifying only the API name) are not valid for this grant type.
-
-To declare a purpose when accessing the CAMARA APIs, it is strongly recommended to guarantee interoperability among Authorization Servers that the `scope` parameter is set to:
-
-`dpv:<dpvValue>#<technicalParameter>`
-
-* `<dpvValue>` is coming from [W3C DPV purpose definition](https://w3c.github.io/dpv/dpv/#vocab-purpose)
-
-* `<technicalParameter>` must be either:
-
-    * one technical scope to limit the request to one API endpoint (this technical scope is described in the API spec YAML file)
-
-    * one API name to cover the complete API (This API name is not described in the API spec YAML file as technical scope) – In this case the request covers all technical scopes of the API.
-
+How to apply the concept of purpose can be found in [Section Purposes of CAMARA APIs access and user consent management](https://github.com/camaraproject/IdentityAndConsentManagement/blob/release-0.2.0/documentation/CAMARA-API-access-and-user-consent.md#applying-purpose-concept-in-the-authorization-request).
 
 ## ID Token
 
@@ -875,7 +856,7 @@ The following Claims are used within the ID Token for all OAuth 2.0 flows used b
 
 * **sub**
 
-  REQUIRED. Subject Identifier. A locally unique and never reassigned identifier within the Issuer for the End-User, which is intended to be consumed by the Client. It MUST NOT exceed 255 ASCII characters in length. The sub value is a case-sensitive string.
+  REQUIRED. Subject Identifier. A locally unique and never reassigned identifier (i.e. a pseudonymous customer reference) within the Issuer for the End-User, which is intended to be consumed by the Client. It MUST NOT exceed 255 ASCII characters in length. The sub value is a case-sensitive string.
 
 
 * **aud**
@@ -928,13 +909,13 @@ ID Tokens MAY contain other Claims. Any Claims used that are not understood MUST
 
 ## Client Authentication
 
-This section defines a set of Client Authentication methods that are used by Clients to authenticate to the Authorization Server when using the Token or Backchannel Authorization Endpoint.
+This section defines the Client Authentication methods that are used by Clients to authenticate to the Authorization Server when using the Token or Backchannel Authorization Endpoint.
 
-These Client Authentication methods are:
+Only one authentication method is allowed:
 
 * **_private_key_jwt_**
 
-  RECOMMENDED. Clients that have registered a public key sign a JWT using that key. The Client authenticates in accordance with JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants [OAuth.JWT] and Assertion Framework for OAuth 2.0 Client Authentication and Authorization Grants [OAuth.Assertions]. At least the Authorization Server MUST support JWS-signed Object with the RS256 algorithm. The JWT MUST contain the following REQUIRED Claim Values and MAY contain the following OPTIONAL Claim Values:
+  REQUIRED. Clients that have registered a public key sign a JWT using that key. The Client authenticates in accordance with [JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants](https://www.rfc-editor.org/rfc/rfc7523.html#section-2.2). At least the Authorization Server MUST support JWS-signed Object with the RS256 algorithm. The JWT MUST contain the following REQUIRED Claim Values and MAY contain the following OPTIONAL Claim Values:
 
     * **iss**
 
@@ -960,16 +941,9 @@ These Client Authentication methods are:
 
       OPTIONAL. Time at which the JWT was issued.
 
-The JWT MAY contain other Claims. Any Claims used that are not understood MUST be ignored.
+  The JWT MAY contain other Claims. Any Claims used that are not understood MUST be ignored.
 
-The authentication token MUST be sent as the value of the [OAuth.Assertions](https://www.rfc-editor.org/info/rfc7521) `client_assertion` parameter.
-
-The value of the [OAuth.Assertions](https://www.rfc-editor.org/info/rfc7521) `client_assertion_type` parameter MUST be `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`, per [OAuth.JWT](https://www.rfc-editor.org/info/rfc7523).
-
-
-* **_client_secret_basic_**
-
-  OPTIONAL. Clients that have received a client_secret value from the Authorization Server MAY authenticate with the Authorization Server using the HTTP Basic authentication scheme as defined in [RFC 2617](https://www.rfc-editor.org/info/rfc2617).
+  The authentication token MUST be sent as the value of the `client_assertion` parameter. The value of the `client_assertion_type` parameter MUST be `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`.
 
 
 ## References
--- a/documentation/CAMARA-OIDC-profile.md
+++ b/documentation/CAMARA-OIDC-profile.md
@@ -75,10 +75,11 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 
 ## Authorization using the Authorization Code Flow
 
+The [OAuth 2.0 Authorization Code](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1) flow involves exchanging an authorization code for a token. Instead of requesting authorization directly from the resource owner, the client directs the resource owner to an authorization server (via its user-agent), which in turn directs the resource owner back to the client with the authorization code.
 
 ### Authentication Request
 
-An Authentication Request is an OpenID Connect Authorization Request that requests that the End-User be authenticated by the Authorization Server.
+An Authentication Request is an [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html) Authorization Request that requests that the End-User be authenticated by the Authorization Server.
 
 The CAMARA profile uses the following OpenID Connect request parameters with the Authorization Code Flow:
 
@@ -386,7 +387,7 @@ Content-Type: application/json
 
 ## Authorization using the OpenID Connect Client-Initiated Backchannel Authentication Flow (CIBA)
 
-The Backchannel Authentication Endpoint is used to initiate an out-of-band authentication of the end-user. This is done by sending an HTTP POST message directly from the Client to the OpenID Provider's Backchannel Authentication Endpoint, using a request defined in the following subsections.
+The [Client-Initiated Backchannel Authentication (CIBA)](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html) flow is used to initiate an out-of-band authentication of the end-user. This is done by sending an HTTP POST message directly from the Client to the OpenID Provider's Backchannel Authentication Endpoint, using a request defined in the following subsections.
 
 Communication with the Backchannel Authentication Endpoint MUST utilize TLS.
 
@@ -703,7 +704,7 @@ Content-Type: application/json
 
 ## Client Credentials Flow
 
-The Client Credentials grant type is used to obtain a 2-legged Access Token that does not represent a user. This grant type can only be used when no personal user data is processed, and it is only a valid option to access the CAMARA APIs for these specific scenarios.
+The [OAuth 2.0 Client Credentials](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) grant type is used to obtain a 2-legged Access Token that does not represent a user. This grant type can only be used when no personal user data is processed, and it is only a valid option to access the CAMARA APIs for these specific scenarios.
 
 
 ### Access Token Request
@@ -836,7 +837,7 @@ Content-Type: application/json
 
 Oauth scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services.
 
-The Open Gateway Authorization Request MUST contain the `scope` parameter, which is a space delimited, case-sensitive list of ASCII strings (scope values). Multiple scope values can be requested simultaneously, subject to the SP being registered to use those “scopes”.
+The Authorization Request MUST contain the `scope` parameter, which is a space delimited, case-sensitive list of ASCII strings (scope values). Multiple scope values can be requested simultaneously, subject to the SP being registered to use those “scopes”.
 
 For OpenID requests, the scope values MUST include `openid`, to indicate that the request is an OpenID Connect request, followed by other values depending on the specific CAMARA services being requested.
 
@@ -976,6 +977,7 @@ The value of the [OAuth.Assertions](https://www.rfc-editor.org/info/rfc7521) `cl
 * [Data Privacy Vocabulary (DPV)](https://w3c.github.io/dpv/dpv/)
 * [E.164 - The international public telecommunication numbering plan](https://www.itu.int/rec/T-REC-E.164-201011-I/en)
 * [OpenID Connect Core 1.0 specification](https://openid.net/specs/openid-connect-core-1_0.html)
+* [OpenID Connect Client-Initiated Backchannel Authentication Flow - Core 1.0](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html)
 * [RFC 2617 - HTTP Authentication: Basic and Digest Access Authentication](https://www.rfc-editor.org/info/rfc2617)
 * [RFC 3986 - Uniform Resource Identifier (URI): Generic Syntax](https://www.rfc-editor.org/info/rfc3986)
 * [RFC 3966 - The tel URI for Telephone Numbers](https://www.rfc-editor.org/info/rfc3966)
--- a/documentation/CAMARA-OIDC-profile.md
+++ b/documentation/CAMARA-OIDC-profile.md
@@ -1,0 +1,1020 @@
+# CAMARA OIDC Profile
+
+## Table of Contents
+<!-- TOC -->
+* [Introduction](#introduction)
+* [Audience](#audience)
+* [Conventions](#conventions)
+* [Authorization using the Authorization Code Flow](#authorization-using-the-authorization-code-flow)
+    * [Authentication Request](#authentication-request)
+    * [Authentication Response](#authentication-response)
+        * [Successful Response](#successful-response)
+        * [Error Response](#error-response)
+    * [Access Token Request](#access-token-request)
+    * [Access Token Response](#access-token-response)
+        * [Successful Response](#successful-response-1)
+        * [Error Response](#error-response-1)
+* [Authorization using the OpenID Connect Client-Initiated Backchannel Authentication Flow (CIBA)](#authorization-using-the-openid-connect-client-initiated-backchannel-authentication-flow-ciba)
+    * [Authentication Request](#authentication-request-1)
+    * [Authentication Response](#authentication-response-1)
+        * [Successful Response](#successful-response-2)
+        * [Error Response](#error-response-2)
+    * [Access Token Request](#access-token-request-1)
+    * [Access Token Response](#access-token-response-1)
+        * [Successful Response](#successful-response-3)
+        * [Error Response](#error-response-3)
+* [Client Credentials Flow](#client-credentials-flow)
+    * [Access Token Request](#access-token-request-2)
+    * [Access Token Response](#access-token-response-2)
+        * [Successful Response](#successful-response-4)
+        * [Error Response](#error-response-4)
+* [The Scope Parameter](#the-scope-parameter)
+* [ID Token](#id-token)
+* [Client Authentication](#client-authentication)
+* [References](#references)
+* [Appendix A. RSA Key Used in Examples](#appendix-a-rsa-key-used-in-examples)
+    * [Appendix A.1. Authorization Server RSA Key](#appendix-a1-authorization-server-rsa-key)
+    * [Appendix A.2. Client RSA Key](#appendix-a2-client-rsa-key)
+<!-- TOC -->
+
+
+## Introduction
+
+Telco network capabilities exposed through APIs provide great value to customers. By simplifying the complexity of telco networks with APIs and making the APIs available across telco networks and countries, CAMARA enables easy and seamless access. To ensure seamless interoperability and increased security, this technical document introduces the OpenID Connect (OIDC) profile, which is specifically tailored for CAMARA APIs.
+
+OpenID Connect is a widely adopted standard for user authentication and authorization in the context of APIs. This document builds on the basic principles of OIDC and outlines CAMARA-specific guidelines to address CAMARA APIs scenarios and use cases.
+
+The CAMARA OIDC profile includes three different authorization mechanisms: Authorization Code Flow, Client Initiated Backchannel Authentication (CIBA), and Client Credentials. Each method is detailed, providing comprehensive insight into the authorization and token request processes, mandatory parameters, successful response content or potential error scenarios.
+
+Key areas covered include:
+
+* Authorization flows:
+
+    * Authorization Code Flow.
+    * Client Initiated Back Channel Authentication (CIBA).
+    * Client Credentials.
+
+* Scope Parameter. Definition of scope parameter format applicable to CAMARA APIs.
+
+* ID Token Content. CAMARA specific details about the content of the Id token.
+
+* Client Authentication. Specifications for client authentication within the CAMARA OIDC profile.
+
+By encapsulating these elements within the CAMARA OIDC profile, this document aims to provide a comprehensive guide for developers and operators, ensuring consistent implementation and adherence to standardized security measures across the CAMARA ecosystem. The defined OIDC profile not only facilitates the integration process, but also serves as a basic framework for developers wishing to leverage the CAMARA APIs while maintaining security and interoperability.
+
+
+## Audience
+
+The target audience for this document is the service/technical departments of operators exposing network functions via standard CAMARA APIs. And applications or client systems that consume CAMARA standard APIs to make use of the operator's network capabilities.
+
+
+## Conventions
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+
+## Authorization using the Authorization Code Flow
+
+
+### Authentication Request
+
+An Authentication Request is an OpenID Connect Authorization Request that requests that the End-User be authenticated by the Authorization Server.
+
+The CAMARA profile uses the following OpenID Connect request parameters with the Authorization Code Flow:
+
+* **scope**
+
+  REQUIRED. A space delimited and a case-sensitive list of ASCII strings for OAuth 2.0 scope values. OpenID Connect requests MUST contain the `openid` scope value. If the `openid` scope value is not present, the ID Token will not be returned. Other scope values MAY be present. Scope values used that are not understood by an implementation SHOULD be ignored. More information about scope values as described in [The Scope Parameter Section](#the-scope-parameter).
+
+
+* **response_type**
+
+  REQUIRED. OAuth 2.0 Response Type value that determines the authorization processing flow to be used, including what parameters are returned from the endpoints used. When using the Authorization Code Flow, this value is `code`.
+
+
+* **client_id**
+
+  REQUIRED. OAuth 2.0 Client Identifier valid at the Authorization Server.
+
+
+* **redirect_uri**
+
+  REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.1). When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
+
+
+* **state**
+
+  RECOMMENDED. Opaque value used to maintain state between the request and the callback. Typically, Cross-Site Request Forgery (CSRF, XSRF) mitigation is done by cryptographically binding the value of this parameter with a browser cookie.
+
+
+* **nonce**
+
+  RECOMMENDED. String value used to associate a Client session with an ID Token, and to mitigate replay attacks. The value is passed through unmodified from the Authentication Request to the ID Token. Sufficient entropy MUST be present in the nonce values used to prevent attackers from guessing values. For implementation notes, see [OpenID Connect Core 1.0 specification](https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes).
+
+* **prompt**
+
+  OPTIONAL. Space-delimited, case-sensitive list of ASCII string values that specifies whether the Authorization Server prompts the End-User for reauthentication and consent. The defined values are:
+
+    * **_none_**
+
+      The Authorization Server MUST NOT display any authentication or consent user interface pages. An error is returned if an End-User is not already authenticated or the Client does not have pre-configured consent for the requested Claims or does not fulfill other conditions for processing the request. The error code will typically be `login_required`, `interaction_required`, or another code defined in [Authentication Error Response Section](#error-response). This can be used as a method to check for existing authentication and/or consent.
+
+    * **_login_**
+
+      The Authorization Server SHOULD prompt the End-User for reauthentication. If it cannot reauthenticate the End-User, it MUST return an error, typically `login_required`.
+
+  The `prompt` parameter can be used by the Client to make sure that the End-User is still present for the current session or to bring attention to the request. If this parameter contains `none` with any other value, an error is returned.
+
+  If an Authorization Server receives a `prompt` value outside the set defined above that it does not understand, it SHOULD ignore it; in practice, not returning errors for not-understood values will help facilitate phasing in extensions using new prompt values.
+
+
+For example, the client directs the user-agent to make the following HTTP request (with line wraps within values for display purposes only):
+
+```http
+GET /authorize?
+response_type=code&
+client_id=s6BhdRkqt3&
+state=af0ifjsldkj&
+nonce=n-0S6_WzA2Mj&
+redirect_uri=https%3A%2F%2Fclient%2Eexample%2Ecom%2Fcb&
+scope=openid%20example-scope HTTP/1.1
+Host: server.example.com
+```
+
+### Authentication Response
+
+
+#### Successful Response
+
+If the resource owner grants the access request, the authorization server issues an authorization code and delivers it to the client by adding the following parameters to the query component of the redirection URI using the `application/x-www-form-urlencoded `format:
+
+* **code**
+
+  REQUIRED. The authorization code generated by the authorization server. The authorization code MUST expire shortly after it is issued to mitigate the risk of leaks. A maximum authorization code lifetime of 10 minutes is RECOMMENDED. The client MUST NOT use the authorization code more than once. If an authorization code is used more than once, the authorization server MUST deny the request and SHOULD revoke (when possible) all tokens previously issued based on that authorization code. The authorization code is bound to the client identifier and redirection URI.
+
+
+* **state**
+
+  REQUIRED if the `state` parameter was present in the client authorization request. The exact value received from the client.
+
+
+The client MUST ignore unrecognized response parameters.
+
+For example, the authorization server redirects the user-agent by sending the following HTTP response (with line wraps within values for display purposes only):
+```http
+HTTP/1.1 302 Found 
+Location: https://client.example.com/cb?
+code=SplxlOBeZQQYbYS6WxSbIA&
+state=af0ifjsldkj 
+```
+
+
+#### Error Response
+
+If the request fails due to a missing, invalid, or mismatching redirection URI, or if the client identifier is missing or invalid, the authorization server SHOULD inform the resource owner of the error and MUST NOT automatically redirect the user-agent to the invalid redirection URI.
+
+If the resource owner denies the access request or if the request fails for reasons other than a missing or invalid redirection URI, the authorization server informs the client by adding the following parameters to the query component of the redirection URI using the `application/x-www-form-urlencoded` format:
+
+* **error**
+
+  REQUIRED. A single ASCII error code from the following:
+
+    * **_invalid_request_**
+
+      The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed.
+
+    * **_unauthorized_client_**
+
+      The client is not authorized to request an authorization code using this method.
+
+    * **_access_denied_**
+
+      The authorization server denied the request (e.g., the authorization server cannot identify the user).
+
+    * **_unsupported_response_type_**
+
+      The authorization server does not support obtaining an authorization code using this method.
+
+    * **_invalid_scope_**
+
+      The requested scope is invalid, unknown, or malformed.
+
+    * **_server_error_**
+
+      The authorization server encountered an unexpected condition that prevented it from fulfilling the request.
+
+    * **_temporarily_unavailable_**
+
+      The authorization server is currently unable to handle request due to a temporary overloading or maintenance of the server.
+
+    * **_interaction_required_**
+
+      The Authorization Server requires End-User interaction of some form to proceed. This error MAY be returned when the `prompt` parameter value in the Authentication Request is `none`, but the Authentication Request cannot be completed without displaying a user interface for End-User interaction.
+
+    * **_login_required_**
+
+      The Authorization Server requires End-User authentication. This error MAY be returned when the `prompt` parameter value in the Authentication Request is `none`, but the Authentication Request cannot be completed without displaying a user interface for End-User authentication.
+
+    * **_account_selection_required_**
+
+      The End-User is REQUIRED to select a session at the Authorization Server. The End-User MAY be authenticated at the Authorization Server with different associated accounts, but the End-User did not select a session. This error MAY be returned when the `prompt` parameter value in the Authentication Request is `none`, but the Authentication Request cannot be completed without displaying a user interface to prompt for a session to use.
+
+    * **_consent_required_**
+
+      The Authorization Server requires End-User consent. This error MAY be returned when the `prompt` parameter value in the Authentication Request is `none`, but the Authentication Request cannot be completed without displaying a user interface for End-User consent.
+
+
+* **error_description**
+
+  OPTIONAL. Human-readable ASCII text providing additional information, used to assist the client developer in understanding the error that occurred. Values for the `error_description` parameter MUST NOT include characters outside the set %x20-21 / %x23-5B / %x5D-7E.
+
+
+* **state**
+
+  REQUIRED if the `state` parameter was present in the client authorization request. The exact value received from the client.
+
+
+`error_description` parameter is not a valid field to take decisions on the developer side, as its value may vary depending on the server implementation, language, format or context. Therefore, the developer should base their logic only on the `error` parameter. Moreover, `error_description` message is not intended to be displayed to the user, but only to help the developer debug the issue. If the developer needs to provide a custom error message to the user, they should use their own logic to map the error code to a suitable message, taking into account the user’s language, tone and level of detail.
+
+
+For example, the authorization server redirects the user-agent by sending the following HTTP response (with line wraps within values for display purposes only):
+
+```http
+HTTP/1.1 302 Found
+Location: Location: https://client.example.com/cb?
+    error=unsupported_response_type&
+    error_description=Unsupported%20response_type%20value&
+    state=af0ifjsldkj
+```
+
+
+### Access Token Request
+
+The client makes a request to the token endpoint, using the HTTP POST method, by sending the following parameters using the application/x-www-form-urlencoded format with a character encoding of UTF-8 in the HTTP request entity-body:
+
+* **grant_type**
+
+  REQUIRED. Value MUST be set to `authorization_code`.
+
+
+* **code**
+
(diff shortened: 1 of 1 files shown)


mhfoo:
Related to the following:
[IdentityAndConsentManagement Issue 104](https://github.com/camaraproject/IdentityAndConsentManagement/issues/104)
[Open-Gateway-Documents Issue 97](https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/issues/97)

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1020 @@
...
+
+The following is a non-normative example of an authentication request (with line wraps within values for display purposes only):
+
+```http
+POST /bc-authorize HTTP/1.1
+Host: server.example.com
+Content-Type: application/x-www-form-urlencoded
+
+scope=openid%20example-scope&
+login_hint_token=ipport%3A80.90.34.2&
```
mhfoo, line 435: typo: should be ```login_hint```, not ```login_hint_token```.

Is there a consideration to use login_hint_token in the future?

garciasolero: Typo fixed.

`login_hint_token` parameter could be used in scenarios with `client_secret_basic` to provide authentication, integrity and  non-repudiation; but we think that these issues can be solved in the same manner by using the `login_hint` parameter, the recommended `private_key_jwt` authentication method and TLS.

This topic has been discussed in previous conversations and several operators were in favor of using `private_key_jwt`.

AxelNennker (changes requested):
First a general remark: I think this document would be easier to read, if it only specifies what decisions Camara made to improve interoperability and security. E.g. repeating all the request parameters  that are already according to the standard makes it hard to see what Camara requires.

Detailed reviews on specific points follow.

documentation/CAMARA-OIDC-profile.md
palmerabollo: Related issue with comments supporting `private_key_jwt` https://github.com/camaraproject/IdentityAndConsentManagement/issues/104#issuecomment-1877218129.

bigludo7:
Thanks @garciasolero for this contribution. May I ask you to add my colleague @sebdewet in the reviewer list (I cannot do it). Thanks

jpengar:
> Thanks @garciasolero for this contribution. May I ask you to add my colleague @sebdewet in the reviewer list (I cannot do it). Thanks

done.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1020 @@
...
+
+Telco network capabilities exposed through APIs provide great value to customers. By simplifying the complexity of telco networks with APIs and making the APIs available across telco networks and countries, CAMARA enables easy and seamless access. To ensure seamless interoperability and increased security, this technical document introduces the OpenID Connect (OIDC) profile, which is specifically tailored for CAMARA APIs.
+
+OpenID Connect is a widely adopted standard for user authentication and authorization in the context of APIs. This document builds on the basic principles of OIDC and outlines CAMARA-specific guidelines to address CAMARA APIs scenarios and use cases.
+
+The CAMARA OIDC profile includes three different authorization mechanisms: Authorization Code Flow, Client Initiated Backchannel Authentication (CIBA), and Client Credentials. Each method is detailed, providing comprehensive insight into the authorization and token request processes, mandatory parameters, successful response content or potential error scenarios.
+
+Key areas covered include:
+
+* Authorization flows:
```
eric-murray, line 51: To define profiles, it is first necessary to establish exactly which standard(s) these profiles apply to. That is not done explicitly in this document (the OpenID CIBA specification is not even referenced). My understanding is that the relevant baseline standards are:
- [OpenID Connect Core 1.0 incorporating errata set 2](https://openid.net/specs/openid-connect-core-1_0.html) for the `authorization_code` grant
- [OpenID Connect Client-Initiated Backchannel Authentication Flow - Core 1.0](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html) for the `urn:openid:params:grant-type:ciba` grant
- [RFC 6749 The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749) for the `client_credentials` grant

These standards should be clearly referenced either here or in the individual sections dealing with each grant type. The individual sections should then highlight the differences between the standard specification and the profile, particularly for parameters that are optional in the standard.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1020 @@
...
+The client makes a request to the token endpoint by adding the following parameters using the `application/x-www-form-urlencoded` format with a character encoding of UTF-8 in the HTTP request entity-body:
+
+* **grant_type**
+
+  REQUIRED. Value MUST be set to `client_credentials`.
+
+
+* **scope**
+
+  OPTIONAL. A space delimited and a case-sensitive list of ASCII strings.
```
eric-murray, line 720: `scope` must be REQUIRED for CAMARA, because all endpoints will have a scope

nickvenezia:
Is this a Developer Profile being created? 
PII should be addressed if this is a profile for consumer.

documentation/CAMARA-OIDC-profile.md
garciasolero: Direct references to the standards have been added in each of the individual sections of the flows.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+
+## Conventions
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+
+## Authorization using the Authorization Code Flow
+
+The [OAuth 2.0 Authorization Code](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1) flow involves exchanging an authorization code for a token. Instead of requesting authorization directly from the resource owner, the client directs the resource owner to an authorization server (via its user-agent), which in turn directs the resource owner back to the client with the authorization code.
+
```
AxelNennker, line 79: ### Cross-Site Request Forgery (CSRF)

Camara choses PKCE for CSRF protection.

Authorization Servers shall require PKCE [RFC7636] with S256 as the code challenge method.

For the Authorization Code flow, Clients shall use PKCE [RFC7636] with S256 as the code challenge method.

Protection provided by state against CSRF is provided by PKCE.
Protection provided by nonce against CSRF is provided by PKCE.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+## Audience
+
+The target audience for this document is the service/technical departments of operators exposing network functions via standard CAMARA APIs. And applications or client systems that consume CAMARA standard APIs to make use of the operator's network capabilities.
+
+
+## Conventions
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
+
```
AxelNennker, line 75: ## Camara security and interoperability best practises

### Network Layer Protections

Regarding network layer protections Camara follows [FAPI 2.0 Security](https://openid.bitbucket.io/fapi/fapi-2_0-security-profile.html#name-network-layer-protections)

---

### Requirements for all endpoints

All TLS connections between web browsers, clients, authorization
servers, and resource servers shall be protected against network
attackers. To this end, clients, authorization servers, and resource
servers

 1. shall only offer TLS protected endpoints and shall establish connections 
    to other servers using TLS;
 1. shall set up TLS connections using TLS version 1.2 or later;
 2. when using TLS 1.2, shall follow the recommendations for Secure Use of Transport Layer Security in [@!BCP195];
 3. should use DNSSEC to protect against DNS spoofing attacks that can lead to
    the issuance of rogue domain-validated TLS certificates; and
 4. shall perform a TLS server certificate check, as per [@!RFC6125].

**NOTE**: Even if an endpoint uses only organization validated (OV) or extended 
validation (EV) TLS certificates, rogue domain-validated certificates can be used 
to impersonate the endpoint and conduct man-in-the-middle attacks. CAA records
[@!RFC8659] can help to mitigate this risk.

### Requirements for endpoints not used by web browsers

For server-to-server communication endpoints that are not used by web
browsers, the following requirements apply:

 1. When using TLS 1.2, servers shall only permit the cipher suites listed in (#tls-12-ciphers).
 2. When using TLS 1.2, clients should only permit the cipher suites listed in (#tls-12-ciphers).
 3. When using the `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256` or `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384` cipher suites, 
 key lengths of at least 2048 bits are required.

#### TLS 1.2 permitted cipher suites {#tls-12-ciphers}

For TLS 1.2, only the following cipher suites shall be used:

  * `TLS_DHE_RSA_WITH_AES_128_GCM_SHA256`
  * `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`
  * `TLS_DHE_RSA_WITH_AES_256_GCM_SHA384`
  * `TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384`

### Requirements for endpoints used by web browsers

For endpoints that are used by web browsers, the following additional
requirements apply:

  1. Servers shall use methods to ensure that connections cannot be
     downgraded using TLS Stripping attacks. A preloaded [@preload] HTTP
     Strict Transport Security policy [@!RFC6797] can be used for this
     purpose.
  2. When using TLS 1.2, servers shall only use cipher suites allowed in
     [@!BCP195].

AxelNennker (changes requested):

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+  REQUIRED. The authorization code received from the authorization server.
+
+
+* **redirect_uri**
+
+  REQUIRED, if the `redirect_uri` parameter was included in the authorization request, and their values MUST be identical.
+
+
+The client MUST authenticate with the authorization server as described in the [Client Authentication Section](#client-authentication).
+
```
AxelNennker, line 272: Authentication between Aggregator and Operator API server MUST use private_key_jwt because private_key_jwt avoids shared secrets.
Authentication between Client and Aggregator or between Client and Operator API server SHOULD be private_key_jwt, but MAY be if the Client is not capable of using private_key_jwt.

AxelNennker (changes requested):
Camara should avoid shared secrets

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+  "access_token": "SlAV32hkKG",
+  "token_type": "Bearer",
+  "refresh_token": "8xLOxBtZp8",
+  "expires_in": 3600,
+  "id_token": "eyJhbGciOiJSUzI1NiIsImtpZCI6IjFlOWdkazcifQ.eyJpc3MiOiJodHRwOi8vc2VydmVyLmV4YW1wbGUuY29tIiwic3ViIjoiYWJiZTBiZDYtYTE4Ny00MmRjLThhYTQtYWI2NWU0NGZkZTYzIiwiYXVkIjoiczZCaGRSa3F0MyIsIm5vbmNlIjoibi0wUzZfV3pBMk1qIiwiZXhwIjoxMzExMjgxOTcwLCJpYXQiOjEzMTEyODA5NzB9.opu_go5CTK3o4BesX7-hChKLvr5oOKzkH_j3D_luWCFPbejZP86tWxJDT_HxrTHvvxBgopLcYLoDOHPCY2RHDvEjyuYvejy93sUQf3YA7O58aPzlSweDbpbjARq45tjEbY5W7MHHiBcg2Az9s2UZOHY6Hpnc5cEIz4S6geD25pSiSGzmLic8kiZtP8SjDEp5vNJedOhjfEp1-xVyOGSujd9pz4jWRNnN61mq0dWwVFwD9pqdXWvyLVcndvEzXTgSLMTX98gd1wykbGef9Ud-1an37_7JFdOB5pROAZDmeLQvfCeaKy6GJF5fftA1aSlC1u-vVDuZP4Ip8zZdM5ShiA"
+}
+```
+#### Error Response
+
+The authorization server responds with an HTTP 400 (Bad Request) status code (unless specified otherwise) and includes the following parameters with the response:
```
sfnuser, line 339:
```suggestion
The authorization server responds with an HTTP 400 (Bad Request) status code (unless specified otherwise) and includes the following within the response body:
```

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1020 @@
...
+
+
+* **client_id**
+
+  REQUIRED. OAuth 2.0 Client Identifier valid at the Authorization Server.
+
+
+* **redirect_uri**
+
+  REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.1). When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
```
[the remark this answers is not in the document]
garciasolero: The inclusion of the http scheme was only intended for testing scenarios in development environments. If you agree, we could make that restriction explicit in the profile.

sfnuser (commented):
Thanks @garciasolero for the proposal and effort.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1020 @@
...
+{
+  "error": "unauthorized_client"
+}
+```
+
+## The Scope Parameter
+
+Oauth scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services.
+
+The Open Gateway Authorization Request MUST contain the `scope` parameter, which is a space delimited, case-sensitive list of ASCII strings (scope values). Multiple scope values can be requested simultaneously, subject to the SP being registered to use those “scopes”.
```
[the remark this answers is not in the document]
garciasolero: Fixed.

documentation/CAMARA-OIDC-profile.md
garciasolero: Fixed.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+
+To declare a purpose when accessing the CAMARA APIs, it is strongly recommended to guarantee interoperability among Authorization Servers that the `scope` parameter is set to:
+
+`dpv:<dpvValue>#<technicalParameter>`
+
+* `<dpvValue>` is coming from [W3C DPV purpose definition](https://w3c.github.io/dpv/dpv/#vocab-purpose)
+
+* `<technicalParameter>` must be either:
+
+    * one technical scope to limit the request to one API endpoint (this technical scope is described in the API spec YAML file)
```
shilpa-padgaonkar, line 860: This restriction to allow just one technical scope should be removed from this profile.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+
+    * one API name to cover the complete API (This API name is not described in the API spec YAML file as technical scope) – In this case the request covers all technical scopes of the API.
```
shilpa-padgaonkar, line 862: If we would like to continue using the wildcard scope, it should be made part of the API documentation in every subproject so that the API documentation can clearly state what it means to use a wildcard technical scope.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+  "access_token": "SlAV32hkKG",
+  "token_type": "Bearer",
+  "refresh_token": "8xLOxBtZp8",
+  "expires_in": 3600,
+  "id_token": "eyJhbGciOiJSUzI1NiIsImtpZCI6IjFlOWdkazcifQ.eyJpc3MiOiJodHRwOi8vc2VydmVyLmV4YW1wbGUuY29tIiwic3ViIjoiYWJiZTBiZDYtYTE4Ny00MmRjLThhYTQtYWI2NWU0NGZkZTYzIiwiYXVkIjoiczZCaGRSa3F0MyIsIm5vbmNlIjoibi0wUzZfV3pBMk1qIiwiZXhwIjoxMzExMjgxOTcwLCJpYXQiOjEzMTEyODA5NzB9.opu_go5CTK3o4BesX7-hChKLvr5oOKzkH_j3D_luWCFPbejZP86tWxJDT_HxrTHvvxBgopLcYLoDOHPCY2RHDvEjyuYvejy93sUQf3YA7O58aPzlSweDbpbjARq45tjEbY5W7MHHiBcg2Az9s2UZOHY6Hpnc5cEIz4S6geD25pSiSGzmLic8kiZtP8SjDEp5vNJedOhjfEp1-xVyOGSujd9pz4jWRNnN61mq0dWwVFwD9pqdXWvyLVcndvEzXTgSLMTX98gd1wykbGef9Ud-1an37_7JFdOB5pROAZDmeLQvfCeaKy6GJF5fftA1aSlC1u-vVDuZP4Ip8zZdM5ShiA"
+}
+```
+#### Error Response
+
+The authorization server responds with an HTTP 400 (Bad Request) status code (unless specified otherwise) and includes the following parameters within the response body:
```
AxelNennker, line 339: Why does this profile need to repeat everything that does not change compared to the original standard? I think this profile should list the differences only. There would be no need having the latest change to this line if this profile would not even contain this line. (the "within"-change)

OIDC mostly points to OAuth2 error response. 
This profile could either point to OIDC token error responses https://openid.net/specs/openid-connect-core-1_0.html#TokenErrorResponse
Or just say that this profile does not change, add or remove any error responses.

The same for successful responses. Let this profile name the specific changes Camara needs for security and interoperability improvements. Otherwise we introduce new ambiguities when re-phrasing what intended to stay the same.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+```
+
+## The Scope Parameter
+
+Oauth scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services.
+
+The Authorization Request MUST contain the `scope` parameter, which is a space delimited, case-sensitive list of ASCII strings (scope values). Multiple scope values can be requested simultaneously, subject to the SP being registered to use those “scopes”.
+
+For OpenID requests, the scope values MUST include `openid`, to indicate that the request is an OpenID Connect request, followed by other values depending on the specific CAMARA services being requested.
+
```
eric-murray, line 843: The rest of the information in this section is already in [CAMARA-API-access-and-user-consent](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md), which should be referenced and not just duplicated here. All that will happen is that the two descriptions will diverge if suggested changes are adopted in this document.

AxelNennker: Yes, I suggest removing contents form this document that is already in standards or in other other Camara documents.

documentation/CAMARA-OIDC-profile.md
palmerabollo: Yes, this was the short-term agreement reached at https://github.com/camaraproject/IdentityAndConsentManagement/issues/32#. It served to move things forward, but nobody really likes it, so we need to work in a solution for the mid-term.

In my view, this first profile should reflect the current status of the agreements, to release a 1.0 version as soon as possible and add further improvements from there, such as a better way to send the "purpose", support for PKCE (also mentioned in this issue), etc. in a 2.0 version.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1020 @@
...
+## The Scope Parameter
+
+Oauth scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services.
+
+The Open Gateway Authorization Request MUST contain the `scope` parameter, which is a space delimited, case-sensitive list of ASCII strings (scope values). Multiple scope values can be requested simultaneously, subject to the SP being registered to use those “scopes”.
+
+For OpenID requests, the scope values MUST include `openid`, to indicate that the request is an OpenID Connect request, followed by other values depending on the specific CAMARA services being requested.
+
+A purpose must be declared within the requested scope for the following grant types:
```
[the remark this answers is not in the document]
palmerabollo: You are right, but a purpose is not the same as a scope. The purpose is the reason why you are accessing the data (e.g. "I want a token to access the 'sim-swap' scope for the purpose 'dpv:FraudManagement'").

As I say in https://github.com/camaraproject/IdentityAndConsentManagement/pull/113/files#r1470035920 the current approach to convey the purpose and its format was a short-term agreement. There were other options on the table (e.g. sending the purpose in a separate "purpose" parameter). We need to work on a better solution.

mengan:
From Mike@t-mobile usa. 

i have several comments on the suggested oidc profile for camara. 


1) I would like to call out that some carrier will/may require some extensions like DPOP.   in which case your non normative examples would be great to include both a Dpop on the token request and a cnf binding to a key in the access_token issued. (though your references are for the id_token the access tokens themselves may (will for us) be in the form of a jwt also. 

2) T-mobile us has done a lot of work in dpop bindings to client instances. (a third party may have a client Server in control of the registered client keys.  but there may be hundreds/thousands/millions of instances of a client (as a SPA web page, or Device present app. )  in which case the Client server (which creates the client assertion)  wishes the Token by bound to a client instance key. (so the client instance can use the issued access token for various service api calls to the carrier).    Hence why calling out an example in the doc would be helpfull.  client server makes the token call with a refer_binding in the client assertion to the client instance, OR the client server hands the client_assertion to the client instance who can dPOP sign the token request itself.   

3) I notice the CIBA flow did not call out Scope or other OIDC elements like nonce as required or optional in the Post to initiate the request.   From our (tmo us) experience, these back channel requests should contain ALL attributes of the oidc request, especially since an oidc id_token is suggested as a result. 

4) for the CIBA back channel flow. T-Mobile experience has shown a need for a user facing context (UI MESSAGING) need. otherwise users may not have enough information to understand what is being asked of them.  this is an important ani phishing element as well.    examples "to enable the xxx app/purchase". at which point the carrier may then dynamically insert copy appropriate to the scopes being requested. (company xyz is asking for access to your device location, Quality of service, make and receive calls and messages, .....  for your number).         Note: an alternative approach is that when a company/client is defined attributes of that client id definition define this user facing copy.   

5) ciba ttl... through experience we (tmo us) have found two different default experiences for ciba. the first what we might call User aware flows could be for when a user is on the phone to a call center, or visiting a retail location.. the user is aware of a transaction and can quickly look at their phone. For these kinds of flows we see a ttl of 5-10 minutes to be appropriate.  The second kind of expereince is where the user is not aware of the transaction, this can even include cases where the user is not even awake or might not have their phone in their hands and charged... these kinds of flows a ttl in the 8-12 hours is typical.   therefore tmo has in the past let clients pass a ttl (number of seconds) as an attribute in the ciba call flow. 


6) your Client credential flow specifically calls out that SCOPES are not needed..  i think this might be a mistake. we believe there will be examples where a client uses user facing (oidc code flow, or ciba flow) for the FIRST interaction with a user. but then once that user has provided the consents... the client could continue (without needing to make refresh token juggling) to simply make a request for a client credential token and request the scope in that token...   Once that token hits the resource service. the service can verify both (1) the client is allowed to hit this service because idp confirmed they have scope for this...  (2) the User the client is trying to hit with this service as a body attribute to the resource api call...  has a recorded consent for this scope to this client.                  Note: in addition to DIRECT user interaction to collect scope consents, T-Mobile usa is also looking at other ownership flows....    Ie... if Company xyz has a billing account with thousands of phone lines, that company  Could create a client and at client provisioning time we can record these "consents".  the client could make an api call to any of the thousand phone numbers and because this client has an exisitng consent record (via an ownership role) the transactions would be processed.   on the other hand if the client tried querying a different number the request may fail with 403 as no consent (via explicit user consent, or ownership role) exists. 


hmm...  overall its a great first stab at getting everyone on the same page for oidc elements for all carriers to support. 

Thanks.

documentation/CAMARA-OIDC-profile.md
shilpa-padgaonkar: Thanks @palmerabollo for your feedback. The current short term agreements are already documented in https://github.com/camaraproject/IdentityAndConsentManagement/blob/release-0.1.0/documentation/CAMARA-API-access-and-user-consent.md and are part of release 0.1 of ICM.  Release 0.2 is now focussed on urgent problems/open issues that arise due to open interpretations of release 0.1. This profile can now focus on improvements. If we are able to get this review completed early with active feedbacks & participation, we could add it as a part of release 0.2. If not, we can bring it in 0.3 release. 
I am not sure if we get any advantage by documenting this profile to reflect again the short term agreements. But we can also get more feedback from others on this.  WDYT?

documentation/CAMARA-OIDC-profile.md
Elisabeth-Ericsson: I suggest to deprecate using the wildcard scope. If used at all, it must be allowed only for APIs which name it in the API documentation.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+
+* _authorization_code_
+
+* _urn:openid:params:grant-type:ciba_
+
+No purpose is required for the `client_credentials` grant type, as this grant type can only be used when no personal user data is processed. The requested scope must be set directly to the value defined for the relevant endpoint within the OAS ("YAML") specification. "Wild card" scopes (i.e. specifying only the API name) are not valid for this grant type.
+
```
Elisabeth-Ericsson, line 851: Even for client_credential, a purpose MUST be sent. An API client must sign up for using a certain API for a certain purpose. The purpose might be charging relevant. If a client has signed up for the same API, but different purposes, it will not be possible to detect the correct monetization / charging rule, unless the purpose is sent.

documentation/CAMARA-OIDC-profile.md
Elisabeth-Ericsson: @shilpa-padgaonkar: fully agree, some text phrases above already allow for multiple scopes, so we must not limit here,

documentation/CAMARA-OIDC-profile.md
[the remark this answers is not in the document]
Elisabeth-Ericsson: I fully agree, the purpose is not the same as a scope. It is the reason for calling an API.
I concur palmerabollo's analysis: let's go for a separate purpose parameter. If then an authorization server does not understand it, it can ignore it.

documentation/CAMARA-OIDC-profile.md
palmerabollo: I agree with supporting private_key_jwt, and only private_key_jwt ("SHOULD ... BUT ..." leaves it open to interpretations and interoperability issues).

I think it is important to omit references to aggregators in CAMARA; this profile should be about API Clients and API Servers.

documentation/CAMARA-OIDC-profile.md
palmerabollo: Thanks @shilpa-padgaonkar. If I get you right, what you suggest is:

- Release 0.2 - Include a first version of this profile focused on urgent problems/open issues that aries due to open interpretations of release 0.1. _(e.g. clearly state possible error codes, required fields that weren't clear, clarify whether private_key_jwt or client_secret_basic should be used since current docs are ambiguous, etc)_
- Release 0.3 - Include further improvements in topics not yet covered _(e.g. PKCE is mentioned in some comments, adherence to FAPI 2.0, etc)._

That sounds good to me. However, I see this particular point (purposes & user consents) as a significant topic that requires some thinking to devise a better solution this time —more than just a patch on top of what we have. That is why I'd like start the discussion as soon as possible but in a separate issue, with the release 0.3 in mind.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+    * **_ipport_**
+
+      For IPv4 and IPv6 addresses, that can optionally include a port. For example, `ipport:80.90.34.2:16790`, `ipport:80.90.34.2`, `ipport:[2001:db8::1]:8080` or `ipport:[2001:db8::1]`.
+
+
+Because in the CIBA flow, the Authorization Server does not have an interaction with the end-user through the consumption device, it is REQUIRED that the Client provides a `login_hint`.
+
+An authentication request is made using the HTTP POST method with the aforementioned parameters in the `application/x-www-form-urlencoded` format and a character encoding of UTF-8 in the HTTP request entity-body. When applicable, additional parameters required by the given client authentication method are also included (e.g. JWT assertion-based client authentication uses `client_assertion` and `client_assertion_type`).
+
+The client MUST authenticate with the authorization server as described in [Client Authentication Section](#client-authentication).
```
eric-murray, line 426: If the client authentication method is `private_key_jwt`, then it should be mandated that the authorisation request be signed (as defined [here](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1.1)). An example of a signed request should be provided.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+The following Claims are used within the ID Token for all OAuth 2.0 flows used by OpenID Connect:
+
+* **iss**
+
+  REQUIRED. Issuer Identifier for the Issuer of the response (the Authorization Server). The iss value is a case-sensitive URL using the https scheme that contains scheme, host, and optionally, port number and path components and no query or fragment components.
+
+
+* **sub**
+
+  REQUIRED. Subject Identifier. A locally unique and never reassigned identifier within the Issuer for the End-User, which is intended to be consumed by the Client. It MUST NOT exceed 255 ASCII characters in length. The sub value is a case-sensitive string.
```
eric-murray, line 878: "Globally unique" end user identifiers, such as MSISDN, IMSI or public IP:port, should never be allowed. It should not be possible to identify the end user from the "sub" field alone.

documentation/CAMARA-OIDC-profile.md
eric-murray: The `purpose` is not intended to support different charging models. If an API provider wishes to charge a given customer different amounts for the same API depending on why the customer is calling that API, then they must do that by allocating different client ids to that customer.

CAMARA should not attempt to standardise "use cases". Whilst an intended use case is important for API design, restricting use cases to a defined set is fatal for any ecosystem claiming to be "open".

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+      OPTIONAL. Time at which the JWT was issued.
+
+The JWT MAY contain other Claims. Any Claims used that are not understood MUST be ignored.
+
+The authentication token MUST be sent as the value of the [OAuth.Assertions](https://www.rfc-editor.org/info/rfc7521) `client_assertion` parameter.
+
+The value of the [OAuth.Assertions](https://www.rfc-editor.org/info/rfc7521) `client_assertion_type` parameter MUST be `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`, per [OAuth.JWT](https://www.rfc-editor.org/info/rfc7523).
+
+
+* **_client_secret_basic_**
```
eric-murray, line 970: `client_secret_basic` should be removed as an option. `private_key_jwt` should be mandatory.

documentation/CAMARA-OIDC-profile.md
garciasolero: As far as I know, the use of the `request` parameter is not approved by all of us. This profile only includes what is already agreed, but we could add it in the future if there is consensus. Anyway, the standard does not say that it is mandatory to pass the `request` parameter if the authentication method is `private_key_jwt`. In fact a few lines above the link you sent there is an example of `private_key_jwt` authentication without request object parameter.

eric-murray: Sure, but if we were only following what the standard says, then we would not need this profile.

The purpose of the profile is to **reduce** the number of options available. But here we still have two - signed or unsigned requests, applicable to all allowed client authorisation schemes. But if the API provider already knows the client public key for `private_key_jwt`, then verifying a signed request is trivial.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+
+## Client Authentication
+
+This section defines a set of Client Authentication methods that are used by Clients to authenticate to the Authorization Server when using the Token or Backchannel Authorization Endpoint.
+
+These Client Authentication methods are:
+
+* **_private_key_jwt_**
+
+  RECOMMENDED. Clients that have registered a public key sign a JWT using that key. The Client authenticates in accordance with JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants [OAuth.JWT] and Assertion Framework for OAuth 2.0 Client Authentication and Authorization Grants [OAuth.Assertions]. At least the Authorization Server MUST support JWS-signed Object with the RS256 algorithm. The JWT MUST contain the following REQUIRED Claim Values and MAY contain the following OPTIONAL Claim Values:
```
eric-murray, line 937: RFC 7521 should not be referenced here because the relevant sections of that document for client authentication are already referenced by RFC 7523. So RFC 7523 is the only inline reference required.

documentation/CAMARA-OIDC-profile.md
garciasolero: That is why it is explained that the sub would be a local unique identifier for authorization server and not reassignable, and the MSISDN or IP could be reassigned.

eric-murray: That explanation is not nearly sufficient. A "globally unique" identifier is also "locally unique", and many MSISDNs have never been re-assigned. What must be avoided is that the API provider just echoes the `login_hint` in the `sub` field. The identifier must be pseudonymous, and this needs to be clearly explained.

But your comment raises another issue. If the identifier identifies an end user (i.e. a person) and always identifies that same user, then the API consumer may learn that a given end user is using a new MSISDN if the `sub` field matches a previous `sub` field obtained for a different MSISDN. This is information that they will not necessarily already know, and would count as a privacy breach. So the identifier should not be constant for a given end user for this reason. An end user who changes their MSISDN should get a new `sub`.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+
+### Authentication Request
+
+An Authentication Request is an [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html) Authorization Request that requests that the End-User be authenticated by the Authorization Server.
+
+The CAMARA profile uses the following OpenID Connect request parameters with the Authorization Code Flow:
+
+* **scope**
+
+  REQUIRED. A space delimited and a case-sensitive list of ASCII strings for OAuth 2.0 scope values. OpenID Connect requests MUST contain the `openid` scope value. If the `openid` scope value is not present, the ID Token will not be returned. Other scope values MAY be present. Scope values used that are not understood by an implementation SHOULD be ignored. More information about scope values as described in [The Scope Parameter Section](#the-scope-parameter).
```
eric-murray, line 88: Either the `openId` scope value is mandatory, or it is not.

If you are defining different behaviour when the `openid` scope value is not present, then that is an additional authorisation scheme whose baseline is the OAuth2 authorisation code flow, and not the OpenId Connect Core authorisation code flow.

So the authorisation code flow section should be separated into two parts:
- Authorisation code flow with `openId` scope value, with OpenId Connect Core as the baseline
- Authorisation code flow without `openId` scope value, with RFC 6749 as the baseline

documentation/CAMARA-OIDC-profile.md
shilpa-padgaonkar: @palmerabollo @jpengar @diegogonmar : I went through the last TSC meeting minutes (audio recording) dated 18th Jan and @diegogonmar as a part of ICM updates has correctly mentioned that 0.2 will focus on solving open issues and 0.3 will include work on profile which will include enhancements. Are we now changing this agreement by saying profile will include contents of 0.1 and possibly 0.2?

documentation/CAMARA-OIDC-profile.md
```diff
@@ -0,0 +1,1022 @@
...
+```http 
+HTTP/1.1 200 OK
+Content-Type: application/json
+```
+```json
+{
+  "access_token": "SlAV32hkKG",
+  "token_type": "Bearer",
+  "refresh_token": "8xLOxBtZp8",
+  "expires_in": 3600,
```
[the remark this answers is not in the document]
garciasolero: The 10-minute recommendation only applies to the authorization code. If you want to raise the debate of token lifetime recommendations, it may be best to open a new issue to discuss it so that it can be included in future versions of the profile.

Elisabeth-Ericsson:
@all: First of all I want to give a big appreciation to the colleagues from Telefonica for putting this great document together. In my view we should find a way to properly review it and get it into the best possible form. This will be the central document used by all developers and thus requires detailed review from all participants. I think that we need time for that.

Given all the discussions in the team, I read the agreement documented in the technical steering group about the release content.  ([2024-01-18 TSC Minutes - CAMARA Project - Confluence](https://wiki.camaraproject.org/display/CAM/2024-01-18+TSC+Minutes))
extracted from the meeting minutes is the following:
_1. V0.1 to be generated. All issues marked as required for the release are solved (Intended date for release will be next Jan 22 - Jan 26.)
2. V0.2 to be progressed as fast as possible. Scope: clarifications on existing scope requested by existing issues
3. V0.3 will include enhancements_
_OIDC profile:
To be created in CAMARA. Not to rely on starting it in OIDF working group to avoid risk of delay._

The agreement does not state the release, the OIDC profile document should be put in.
In order to buy us some discussion time, would it be possible to include the most needed clarifications on the short term agreement in the already existing document (https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md) as 0,2 - and only concentrate on the most urgent problems only ? In parallel we would safe time and get some room for discussions on the new document for the OIDC profile?
Then we would use the new document on OIDC profile (content of this pull request) to be the embryo of 0,3. I acknowledge that the ODIC profile document is very much needed. But is it reasonable to spend so much effort in this (very good) document and have it focused on the short term agreement (which is not really future proof) ?

I'm a bit afraid that this new document will become the bible. As soon as the implementations will take off now, gain speed and then - with this short term agreement - we cement it and build a burden into our solution we cannot get rid of any more.

documentation/CAMARA-OIDC-profile.md
diegogonmar: Hi @shilpa-padgaonkar. Activity in TSC was reporting on Identity & Consent group relevant progress, not escalation to taking any decision there. Decission in Identity&Consent was to use v0.2 for clarifications and urgent issues, I think that is clear. In parallel we decided to write OIDC profile in CAMARA. Then, once first version was proposed, as part of this github thread discussion and as per yesterday call, it was proposed to use a first version of OIDC profile as an appropriate way to cover current picture and such urgent issues targeted for v0.2. This is indeed a proposal as it could be the most effective way to solve points that need to be clarified and the urgent issues, in an organized manner. We had some interesting discussions yesterday and you weren't opposing to have profile already in v0.2, not sure if you are objecting to that now, as you sent this comment after yesterday call

documentation/CAMARA-OIDC-profile.md
garciasolero: Changed to REQUIRED.

documentation/CAMARA-OIDC-profile.md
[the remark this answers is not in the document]
garciasolero: Added restriction to support http schema in non-production environments only.

documentation/CAMARA-OIDC-profile.md
garciasolero: Removed duplicates from internal documentation. Right now the link to that information is broken, the new reference link will work when the new version (0.2.0) is created in the repository.

documentation/CAMARA-OIDC-profile.md
eric-murray: > the `request` object parameter will be redundant for security purposes

No, not redundant. Signed authorisation requests ensure the integrity of the request, and prevent it being modified by a MITM (or, more accurately, allow such modifications to be detected).

Now it could be argued that TLS will prevent such attacks and thus render them purely theoretical. But nonetheless signed requests are defined as an option by CIBA and, given that there is no good reason not to support signed requests if the client public key is anyway known, will likely be mandated by at least some API providers.

So I think signed authorisation requests should be mandatory (given that `client_secret_basic` is no longer being proposed as a valid client authentication scheme). That would remove another CIBA option, thus increasing the likelihood of the API provider and consumer supporting a compatible authorisation flow.

documentation/CAMARA-OIDC-profile.md
garciasolero: The reference to client_secret_basic as a supported authentication method is removed, as agreed at the last meeting.

documentation/CAMARA-OIDC-profile.md
garciasolero: The links have been changed to point to the new RFC 7523 reference.

documentation/CAMARA-OIDC-profile.md
garciasolero: I have modified the text to indicate that the `sub` claim should be a pseudonymous customer reference.

documentation/CAMARA-OIDC-profile.md
jpengar: Please see https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#discussion_r1474604755

documentation/CAMARA-OIDC-profile.md
jpengar: As per yesterday's WG meeting call (31/01), after asking the WG participants to consider the different opinions presented during the meeting, the way forward was agreed to go with the self-contained format + compromise to properly note any changes related to the reused standard.

jpengar:
> During the meeting there was the argument that a repeat-everything profile would be easier to read for developers. I do not think that argument is relevant because nobody sane is going to start developing e.g. an AZ from scratch. OIDF recommends to start from scratch but to use existing libraries. https://openid.net/developers/certified-openid-connect-implementations/
> 
> Most of the telcos either already have an OIDC implementation and if not we pick a vendor who has some proven track record implementing OIDC and OAuth2 SW.
> 
> So, I think that this policy should list security requirements that improve the security of OIDC. And this policy should reduce the number of features an implementer has to implement fostering interoperability. Example: if we require PKCE to be used in authorization code flow as an Anti-CSRF then developers know that other measures to counter CSRF are not needed. Example: if we require sender-constraint access tokens with DPoP then developers know what to do.
> 
> If this profiles repeats all the standards it profiles than this document gets quite long, which is then unreadable for developers. So, I think that "readability" is not an argument that holds water.

I understand your point, and you presented it correctly in yesterday's call, but after asking the WG participants and considering the different opinions presented during the meeting, the way forward was agreed to go with the self-contained format + compromise to properly note any changes related to the reused standard.

documentation/CAMARA-OIDC-profile.md
```diff
@@ -100,7 +100,7 @@ The CAMARA profile uses the following OpenID Connect request parameters with the
 
 * **redirect_uri**
 
-  REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.1). When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
+  REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986.html#section-6.2.1). When using this flow, the Redirection URI MUST use the https scheme; however, it MAY use the http scheme only for non-production environments. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
```
AxelNennker, line 103: Please keep the text of the original standard, or delete this definition because it has not changed.
I think, this profile should not care what developers do in non-production. Also there are ways to have certificates on internal servers and developer machines e.g. self-signed certs and /etc/host entries.

AxelNennker (changes requested):

jpengar:
@Elisabeth-Ericsson Regarding your comment https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#issuecomment-1921382449, please see Diego's comment here -> https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#discussion_r1474604755

And regarding your concerns, the existing purpose agreement is what's going to be implemented in the short term anyway. It is not more or less valid/official because we documented it in CAMARA-API-access-and-user-consent.md and/or the OIDC profile. And as I said yesterday, the OIDC profile is not only about the purpose solution, but also about including other implementation details that have already been requested as needed by other participants in their current implementations (client authentication, CIBA login_hint, error scenarios, etc.). That's why the profile is proposed for the v0.2.0 release. And in fact, in the latest updates of the profile in this PR, the purpose solution is just a reference to API-access-and-user-consent.md.

documentation/CAMARA-OIDC-profile.md
eric-murray: Sure, but signed or unsigned authorisation requests both remain as options. One of those options should be removed. I propose to mandate signed requests.

jpengar:
> Yesterday I presented a different point. This point is about the "readablity"-argument, which yesterday I had no counter argument to, but today I have the above and I think that the readability argument is wrong.
> 
> I think the changes-only policy document is much easier to read.
> 
> Usually, in this working group we discuss arguments for a few days and give participants time to review and comment.
> 
> See the redirect_uri example: Today a change was merged, that I think is wrong and it also does not mark the definition as changed from the standard.

There have been days of feedback on the profile format since the PR was created, there have been different opinions, and yesterday's meeting was precisely to agree on a way forward and unblock this hot topic.

If you've thought about it a second time and now you've come up with some more analysis, that's fine. But yesterday we did agree on a way forward. So I understand that you are now proposing to reopen the discussion?? 

If the rest of the participants feel it's necessary, I would disagree, but it would be the WG decision. And in that case we can give some time for feedback as we always do. But this has a big impact on the PR, and we were already working and updating the PR according to yesterday's agreement. It is an inconvenience.

documentation/CAMARA-OIDC-profile.md
garciasolero: @eric-murray , you are using the MITM argument to refute my claim that it could be used in the same way to not trust the public keys exposed in the jwks_uri.

documentation/CAMARA-OIDC-profile.md
garciasolero: You are precisely saying that it should be referenced because there are no changes when in fact there are, or we want there to be. In another thread in this PR you said that the use of the `https` schema should be mandatory.

This is exactly what the [OIDC standard says](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest):

_redirect_uri
    REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in Section 6.2.1 of [[RFC3986]](https://openid.net/specs/openid-connect-core-1_0.html#RFC3986) (Simple String Comparison). **When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme, provided that the Client Type is confidential, as defined in Section 2.1 of OAuth 2.0, and provided the OP allows the use of http Redirection URIs in this case. Also, if the Client is a native application, it MAY use the http scheme with localhost or the IP loopback literals 127.0.0.1 or [::1] as the hostname.** The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application._

And what exactly the [OAuth2.0 standard says](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.1):

**This specification does not mandate the use of TLS** because at the time of this writing, requiring clients to deploy TLS is a significant hurdle for many client developers.

documentation/CAMARA-OIDC-profile.md
eric-murray: Well, if a MITM is intercepting **all** traffic between the API consumer and provider at HTTP level, then indeed that client is hopelessly compromised. But that is an even more theoretical attack. I think we have to make the assumption that the API provider knows the correct client public key.

But the **key point** remains that this profile allows both signed and unsigned authorisation requests. Only one of these options should be allowed, and my preference is signed authorisation requests. As long as signed authorisation requests remain an option, some API providers will mandate it because their own internal security rules will not allow accepting unsigned requests when they could be signed.

garciasolero: I am not particularly keen on rejecting the inclusion of the `request` parameter in the CIBA flow. We can include this requirement in future versions of the profile if there is consensus. 

Nice theoretical discussion. :smiley:

AxelNennker:
I created https://github.com/AxelNennker/IdentityAndConsentManagement/blob/camara_oidc_profile/documentation/CAMARA-Security-Interoperability.md

Much shorter much easier to read, I think.
Hopefully I captured everything that CAMARA-OIDC-Profile wants to say about usage of OIDC, CIBA, etc

documentation/CAMARA-OIDC-profile.md
AxelNennker: What I meant with "don't change the standard" is that the text of a parameter definition should be verbatim the same as in the standard. 
We agreed in Camara that we use OIDC authorization code flow, so the OIDC standard is the basis of this PR.

Regarding "at this time":
> This specification does not mandate the use of TLS because at the time of this writing, requiring clients to deploy TLS is a significant hurdle for many client developers.

[OpenId Connect is now ten years old](https://self-issued.info/?p=2468) and OAuth2 even older. A lot has changed and improved.

The document is intended to be a profile of OIDC not a redefinition of OIDC based on OAuth2.
Please keep the OIDC definitions of parameters, if possible. If you change a definition, then please clearly mark the change.

I still think that just stating the differences to the profiled standards leads to a document that is easier to read and better for implementors. Please see https://github.com/AxelNennker/IdentityAndConsentManagement/blob/camara_oidc_profile/documentation/CAMARA-Security-Interoperability.md

nickvenezia: @AxelNennker brings up a lot of great points. I see a major issue with this and the demand for Data Minimization. @AxelNennker did you come across issues with the OIDC Profile and Data Minimization Regulation in your research?

documentation/CAMARA-OIDC-profile.md
hdamker: > Please see [#113 (comment)](https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#discussion_r1474604755)

@jpengar can you shortly explain what the referred comment has to do with the content of this review comments from @shilpa-padgaonkar and @Elisabeth-Ericsson here (way to handle wildcard scope?).

jpengar: > @jpengar can you shortly explain what the referred comment has to do with the content of this review comments from @shilpa-padgaonkar and @Elisabeth-Ericsson here (way to handle wildcard scope?).

@hdamker Long story short, my comment was actually more related to Eli's comment (https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#discussion_r1471453854) and I just wanted to highlight why the "wildcard" scope is present in the document. My bad, I didn't make my point clear enough :/

So, my point in including that reference comment was to highlight that the reason the current profile content includes the agreed upon "wildcard" (api name, bundle scope... people call it different things xD) is because this PR proposes to use a first version of the OIDC profile as an appropriate way to cover the current state (a.k.a. short-term agreement) and urgent issues/clarifications targeted for v0.2. But of course it is still pending to discuss the new/improved solution to handle purpose and it would be properly documented in v0.3. And if decided, remove the use of wildcard scopes, etc.

hdamker:
@garciasolero @jpengar @Elisabeth-Ericsson @shilpa-padgaonkar @eric-murray 

First of all thanks to @garciasolero for the initial draft of this document. I understand that it is written with two different audiences in mind

> The target audience for this document is the service/technical departments of operators exposing network functions via standard CAMARA APIs. And applications or client systems that consume CAMARA standard APIs to make use of the operator's network capabilities.

But despite the great effort I think that it is unfortunately missing both audiences with the attempt to be "self contained".

Let me start with "service/technical departments of operators exposing network functions via standard CAMARA APIs". They would rely on the referred standards and need to know (only) the points which CAMARA is defining in addition to ensure interoperability and security. For them the document has too much duplicated information which is already defined with in the referenced standards. They would need exactly the information which is provided within the [short profile document](https://github.com/AxelNennker/IdentityAndConsentManagement/blob/camara_oidc_profile/documentation/CAMARA-Security-Interoperability.md) proposed by @AxelNennker. For the exact definition of the expected behaviour the document in this PR has a high risk to get an "alternative standards" to OIDC, but without being ever complete (e.g. the complete definition of terminology is missing). 

On the other side are the "applications or client systems that consume CAMARA standard APIs". Actually the audience are here the developers of the these applications or client systems. For them indeed a self-contained document which explains what they need to initiate the flows, authenticate, what the expect in the tokens, explanations of potential errors, etc would be helpful. We should not require them to read multiple documents (also not within CAMARA) and to dig into standards which are written for implementators. But that should be a documentation from a developer perspective, not mixed up with design decision how to implement the OIDC standards. Reading over the current document I don't have the impression that the current PR would fulfil the need of this audience.

Said this my recommendation would be to **reduce the profile document in this PR to just the differences to the referred standards** and consider to write an additional document which describes with examples, flow diagrams etc the use of CAMARA APIs from developers perspective, referring to the standards for further details, including mentioning the differences / limitations imposed by the CAMARA profile to ensure interoperability. This approach would ensure that we can fix in short time the necessary agreements for implementations and then in parallel to implementations work on proper documentation for developers (the main target audience of CAMARA APIs).

documentation/CAMARA-OIDC-profile.md
diegogonmar: Hi @hdamker , thanks for your view. I didn't say anything about Shilpa referring to an escalation, I just added that as an information.
 
I think you say something correct: _the profile was planned for 0.3.0_. Then please read again my comment above: _as part of this github thread discussion and as per yesterday call, it was proposed to use a first version of OIDC profile as an appropriate way to cover current picture and such urgent issues targeted for v0.2_. Thus, I think its clear and I hope its fine, that this group opened a discussion on whether to include a first version of the profile for v0.2 was the best way to reflect urgent issues. 
It's not in conflict with existing agreement of having an OIDC profile for v0.3, which of course as agreed will include enhancements. So after last wednesday call you challenge if having a profile already in v0.2 it's appropriate, or at least you consider it adds no benefit. 

I'd close this part of the discussion and focus on way forward. Regarding your last paragraph, it's a valid suggestion (not saying I agree with it), can you put it in separate thread comment so we collect group opinion on that suggestion?

hdamker:
@diegogonmar asked that I put my following suggestion from a now closed discussion into a separate thread comment so we collect group opinion on it (origin: https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#discussion_r1478838189):

> As a comprise I would propose that the profile can be already part of v0.2, if in time agreed, and if it does not codify the short-comings of the short-term solution. Which includes that it also not refer to (https://github.com/camaraproject/IdentityAndConsentManagement/blob/release-0.2.0/documentation/CAMARA-API-access-and-user-consent.md#applying-purpose-concept-in-the-authorization-request) but only the other way round (the API-access-and-user-consent.md refers for additional clarifications to the profile).

I would like to see it together with my proposal within the previous [comment](https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#issuecomment-1927764400), to focus first on a short profile document to get the necessary clarifications agreed, and doing a comprehensive developer documentation afterwards based on the profile.

documentation/CAMARA-OIDC-profile.md
hdamker: @diegogonmar I agree to close this discussion and look forward. Have followed your proposal and created https://github.com/camaraproject/IdentityAndConsentManagement/pull/113#issuecomment-1940905690.

davidgtonge:
Just adding to the discussion, there is quite a lot of repetition here from existing specs. Is it worth considering referencing https://openid.bitbucket.io/fapi/fapi-2_0-security-profile.html 

This is a secure profile of OAuth2 that has the following advantages:
1. extensive conformance tests
2. extensive vendor support
3. formal security analysis

Axel's PR looks like it could build on this existing work.

AxelNennker:
Regarding offline_access, please mark this as a change to the OIDC standard on [offline_access](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess).
The standard defines that the client gets an access token for the user-info endpoint.
While this profile defines that the clients gets a refresh token and and access token for a RS endpoint

documentation/CAMARA-OIDC-profile.md
AxelNennker: @garciasolero the definition of redirect_uri is different to the standard e.g. the part about confidential clients is not here instead there is the mention of non-production environments.

I read this definition as being more restrictive than the OIDC standard and I support that https is used for all client types and url-schemes for native apps. But maybe this is not the only interpretation. 
I think, changes to the standard should be clearly marked.

AxelNennker (changes requested):
Please mark changes to the standards

garciasolero:
This self-contained profile pull request is being closed because it has been decided to use another profile format that has been merged into `main` branch in pull request [#121](https://github.com/camaraproject/IdentityAndConsentManagement/pull/121).

PR #114, opened 2024-01-29
Add missing response errors in CIBA flow from GSMA doc

jpengar:
#### What type of PR is this?

* documentation

#### What this PR does / why we need it:

Add some response errors in CIBA documentation according to [specification](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html). This PR replicates minor changes in the GSMA doc merged in https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/pull/86 that have not yet been propagated to CAMARA.

#### Which issue(s) this PR fixes:

Partially fixes / related to Issue #82.

#### Special notes for reviewers:

This does not add anything "new", it is just intended to make sure that there is no information documented in GSMA that is missing in CAMARA. Mainly because the GSMA doc is being adapted to reference the CAMARA doc as per #82, and CAMARA must now include all required information.

Full error information will be detailed in any case in CAMARA OIDC profile: https://github.com/camaraproject/IdentityAndConsentManagement/pull/113

#### Changelog input

```
Add missing CIBA flow error descriptions from GSMA doc.
```

#### Additional documentation 

N/A

PR #121, opened 2024-02-12
Camara OIDC profile

AxelNennker:
#### What type of PR is this?

Add one of the following kinds:
* documentation

#### What this PR does / why we need it:

By restricting options the OIDF and IETF standards offer this document improves CAMARA security and interoperability.
The CAMARA Security and Interoperability document follows the path [FAPI 2.0 Security](https://openid.bitbucket.io/fapi/fapi-2_0-security-profile.html) took, leading to a concise description which helps implementers focusing on those parts of the standards that are required or recommended.

#### Which issue(s) this PR fixes:

Fixes #78 
Fixes #82
Fixes #87 
Fixes #90 
Fixes #100 
Fixes #104  
Fixes #127 
Fixes #132 
Fixes #133 
Fixes #136
Fixes #137

Issue #70, opened 2023-10-04
Front-end flow PII leakage?


chrishowell:
As the consent check is within the unauthenticated auth code stage of the Authorisation Code flow, rather than in the access token stage, wouldn't it be possible to find out whether the user had consented to an operation or even had ever accessed another app? That would constitute PII leakage.

Could the consent check be moved into the access token stage, to avoid this?

jpengar:
As per Yesterday's working group call (04/10), the issue will be discussed in the context of the GSMA technical stream, and anything that needs to be piggybacked into CAMARA will be raised later for everyone in CAMARA to review.

chrishowell:
@jpengar are you going to bring the changes back into this project?

chrishowell:
I'm intrigued - is consent meant to be collected in the authorisation code phase - so the access token wouldn't be obtainable until consent had been collected?
If that's the case, as an attacker I could trigger consent without authentication for all the ASP IDs :|

palmerabollo:
Hi @chrishowell, this topic is actively being discussed in GSMA. You raised this issue regarding front-end flow PII leakage at https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/issues/53, and there is a pull request with an improvement at https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/pull/61 waiting for your input.
 
There is also this one at https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/issues/52, which is also related to the unauthenticated front-end flow. There is a proposal to optionally support RFC9101 (JWT-Secured Authorization Request) also waiting for your feedback.
 
The authorization code flow is standard, i.e. you can also trigger consent in a Google/Facebook/Twitter/... service.

chrishowell:
I am not happy about this being discussed in the closed OpenGateway repo, authentication is a key part of CAMARA and needs to be discussed here too.

I have raised the issue in the open PR:
https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/pull/61/files#r1377367639

palmerabollo:
I agree it would be great to open the GSMA repository to avoid duplicating issues. @MarkCornall, have you considered it?

In my opinion, we should define APIs in CAMARA and handle the integration with aggregators, operators, and other cross-cutting concerns, such as the authentication flows, in GSMA, where a working reference implementation exists.

The worst-case scenario is having the same issues duplicated in both repositories.

PR #123, opened 2024-02-13
Add missing offline access/refresh token doc from GSMA

### LINKED ISSUE #82: Redundancy with GSMA API General Federated Call Flows document
**Problem description**
The document https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md is very helpful but it contains a lot of redundant information with GSMA document: https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/blob/045689e0c42478cb2e10f2829c0cbdeac8a9a184/Chapters/Chapter%2006%20proposal%201.md#api-general-federated-call-flows

 It's probably not a good idea to have 2 technical document on same topic. It will be inconsistency-prone & bring confusion.

**Possible evolution**
Have one document covering the API flows and it will be the reference.

Perhaps the CAMARA document should be expunge from the detailed flow description and refer the GSMA document and only focus around purpose & consent stuff?

**Alternative solution**

**Additional context**
Same comment raised in GSMA: https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/pull/86

jpengar:
#### What type of PR is this?

* documentation

As discussed and agreed in the CAMARA Identity & Consent Management Working Group (meeting call 17/01/24), and as per CAMARA TSC (Technical Steering Committee) decision, the documentation of the NBI CAMARA API interface, authZ/authN flows, purpose/consent management, etc... will only be placed in CAMARA. CAMARA can be the source of truth for northbound interfaces. GSMA Opengateway will simply adopt what is defined in CAMARA.

This PR is intended to move the GSMA Opengateway [Chapter 04 Offline Access.md](https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/blob/main/Chapters/Chapter%2004%20Offline%20Access.md) to CAMARA as a last step to fix #82, taking into account PR #114 as well.

Fixes #82

Please, I strongly suggest that we first focus on moving the necessary documentation to the CAMARA Identity & Consent WG repository as agreed. To have all the documentation here. And then, if the WG participants want to revisit the content of the documentation, we can open a new issue for that, no problem. This would allow us to wrap up a long discussion, close issue #82, and move on to the next needed discussion without blocking this track by creating new dependencies between issues.

#### Changelog input

```
 offline access / refresh token flows
```

#### Additional documentation 

N/A

### DIFF
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -19,6 +19,10 @@ This document defines guidelines for telco operator exposure platforms to manage
       - [Authorization code flow (Frontend flow)](#authorization-code-flow-frontend-flow)
       - [CIBA flow (Backend flow)](#ciba-flow-backend-flow)
       - [Client Credentials](#client-credentials)
+  - [Offline Access - Refresh Token](#offline-access---refresh-token)
+    - [Request a Refresh Token in Frontend Based Flow](#request-a-refresh-token-in-frontend-based-flow)
+    - [Request Refresh Token in Backend Based Flow](#request-refresh-token-in-backend-based-flow)
+    - [Refresh Token Flow](#refresh-token-flow)
   - [CAMARA API Specification - Authorization and authentication common guidelines](#camara-api-specification---authorization-and-authentication-common-guidelines)
     - [Use of openIdConnect for `securitySchemes`](#use-of-openidconnect-for-securityschemes)
     - [Use of `security` property](#use-of-security-property)
@@ -163,7 +167,7 @@ alt Standard OIDC Auth Code Flow between Invoker and API Exposure Platform
     ExpO-->>FE: 302<br>Location: invoker_callback?code=Operatorcode
   else If Consent is NOT granted - Consent Capture within AuthCode Flow  
     Note over FE,ExpO: Start user consent capture process<br>following Section 3.1.2.4 of the OIDC Core 1.0 spec.    
-    ExpO-->>FE: 302<br>Location: aggregator_callback?code=Operatorcode 
+    ExpO-->>FE: 302<br>Location: invoker_callback?code=Operatorcode 
   end
   FE-->>-BE: GET invoker_callback?code=OperatorCode
   BE->>ExpO: POST /token<br> code=OperatorCode
@@ -372,6 +376,186 @@ If some usecase/s for an API point to off-net scenarios and where consumption an
 The Client Credentials grant type is used to obtain a 2-legged access_token that does not represent a user. This grant type can only be used when no personal user data is processed, and it is only a valid option to access the CAMARA APIs for these specific scenarios. 
 
 More details about the standard flow can be found in the official IETF specification [The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4).
+
+## Offline Access - Refresh Token
+
+The defined solution uses Refesh Token as a mean to enable CAMARA API access in Offline Scenarios. Online Access means that there is a Telco-based Network Connection, either Mobile Network connection in mobile use cases or Fixed Network connection in fixed use cases.
+
+- Refresh Token may be issued as result of AuthN/AuthZ, if requested by API invoker and allowed by use case. 
+  - [RFC6749 Section 1.5](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5): _A refresh token is a string representing the authorization granted to  the client by the resource owner.  The string is usually opaque to the client. The token denotes an identifier used to retrieve the authorization information. Therefore, the Refresh Token is bound to the authorization of the end user to the Application._
+
+- As indicated by [OIDC Spec](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess ), one of the uses of refresh token is allowing offline access. 
+
+- Thus, a Refresh Token can be used for Offline Access as it represents an existing authorization, gathered in a previously Online Access. 
+
+- Also, can be used even in Online Access, just to avoid a re-authentication. This would be use case driven and depending on the security needs for each API, e.g.: APIs may require more frequent Auth or even Auth for each use (like Number Verification API). 
+
+This chapter includes the flows for CAMARA showing the usage of Refresh Token.
+
+### Request a Refresh Token in Frontend Based Flow
+
+
+```mermaid
+sequenceDiagram
+autonumber
+title Request Refresh Token - Authorization Code Flow
+participant FE as Device App
+participant BE as Invoker<br>(App Backend/Aggregator)
+box Operator
+  participant ExpO as API Exposure Platform<br>@Operator
+  participant Consent as Consent Master<br>@Operator
+end
+
+Note over FE,BE: Use Feature needing<br>Operator Capability
+Note over BE: Offline Access wanted,<br>thus offline_access scope requested  
+BE->>FE: Auth Needed - redirect <br>/authorize?response_type=code&client_id=coolApp<br>&scope=dpv:<purposeDpvValue>#35;<technicalParameter> offline_access<br>&redirect_uri=invoker_callback...
+FE->>+FE: Browser /<br> Embedded Browser
+alt Standard OIDC Auth Code Flow between Invoker and API Exposure Platform
+  FE-->>ExpO: GET /authorize?response_type=code&client_id=coolApp<br>&scope=dpv:<purposeDpvValue>#35;<technicalParameter> offline_access<br>&redirect_uri=invoker_callback...
+  rect rgb(0, 76, 153)
+    Note over ExpO: Authorization process details included in<br>the "Consume a CAMARA API<br>- Authorization Code Grant (FrontEnd)" flow...
+  end    
+  ExpO-->>FE: 302<br>Location: invoker_callback?code=Operatorcode
+  FE-->>-BE: GET invoker_callback?code=OperatorCode
+  BE->>ExpO: POST /token<br> code=OperatorCode
+  ExpO->>ExpO: Create and save<br>Operator Refresh Token
+  ExpO->>BE: 200 OK <br> {OperatorAccessToken}<br>{OperatorRefreshToken}
+end
+```
+
+**Key concepts**
+
+- Same considerations that apply for general call flows in [Authorization flows / grant types](#authorization-flows--grant-types) also apply for this flow.
+-  Refresh Token issuance is achieved by requesting `offline_access` scope. 
+
+**Flow description** 
+
+The API invoker (which could potentially be the application backend, an aggregator, etc.) instructs the application frontend in the device to initiate the OIDC authorization code flow with the operator as described in the [Authorization code flow (Frontend flow)](#authorization-code-flow-frontend-flow).   
+
+The API invoker includes `offline_access` scope to signal that a Refresh Token is expected (step 1). 
+
+Then same flow depicted in [Authorization code flow (Frontend flow)](#authorization-code-flow-frontend-flow) occurs (steps 2-8). 
+
+> **NOTE: Flow diagram is simplified, full details offered in general call flows** 
+
+When Operator is going to issue the Access Token, it also issues a Refresh Token (OperatorRefreshToken), as `offline_access` scope was requested. 
+
+> NOTE: **Depending on the use case this will be allowed or not.** i.e.: not every API invoker will have by default the right to have a Refresh Token.
+
+Flow will be completed normally, so Operator provides API invoker with both Access Token and Refresh Token (step 8). 
+
+
+### Request Refresh Token in Backend Based Flow 
+
+ 
+```mermaid
+sequenceDiagram
+autonumber
+title Request Refresh Token - CIBA Flow
+participant FE as Device App<br> (Consumption Device)
+participant BE as Invoker<br>(App Backend/Aggregator)  
+box Operator
+  participant ExpO as API Exposure Platform<br>@Operator  
+  participant Consent as Consent Master<br>@Operator
+end
+
+Note over FE,BE: Feature needing<br>Operator Capability  
+Note over BE: Offline Access wanted,<br>thus offline_access scope requested
+Note over BE: Select User Identifier:<br> Ip:port / MSISDN / other TBD  
+
+alt OIDC Client-Initiated Backchannel Authentication (CIBA) Standard Flow between Invoker and Operator.
+  BE->>+ExpO: POST /bc-authorize<br> Credentials,<br>scope=dpv:<purposeDpvValue>#35;<technicalParameter>" offline_access,<br>login_hint including User Identifier    
+  rect rgb(0, 76, 153)
+    Note over ExpO: Authorization process details included in<br>the "Consume a CAMARA API - CIBA flow"...
+  end  
+  ExpO->>BE: HTTP 200 OK {"auth_req_id": "{OperatorAuthReqId}"
+  loop Invoker polls until consent is granted or until expires. If granted in advance, token returned in first poll
+    BE->>+ExpO: POST /token <br>Credentials}<br>auth_req_id={OperatorAuthReqId}
+    ExpO->>ExpO: Create and save<br>Operator Refresh Token    
+    ExpO->>-BE: HTTP 200 OK <br>{OperatorAccessToken}<br>{OperatorRefreshToken} 
+  end  
+end
+```
+
+**Key concepts**
+
+- Same considerations that apply for general call flows in [Authorization flows / grant types](#authorization-flows--grant-types) also apply for this flow. 
+- Refresh Token issuance is achieved by requesting `offline_access` scope 
+ 
+**Flow description**
+
+The API invoker (which could potentially be the application backend, an aggregator, etc.) requests a 3-legged access token to the operator API exposure platform. The process follows the OpenID Connect Client-Initiated Backchannel Authentication (CIBA) flow as described in [CIBA flow (Backend flow)](#ciba-flow-backend-flow).  
+
+The API invoker selects an Identifier and performs `/bc-authorize` call including `offline_access` scope to signal that a Refresh Token is expected (step 1). Then same flow depicted in [CIBA flow (Backend flow)](#ciba-flow-backend-flow) occurs (steps 1-5). 
+
+> **NOTE: Flow is simplified, full details offered in General call flows.** 
+
+When Operator is going to issue the Access Token, it also issues a Refresh Token (OperatorRefreshToken) (step 4), as `offline_access` scope was requested. 
+
+> NOTE: **Depending on the use case this will be allowed or not.** I.e.: not every API invoker will have by default the right to have a Refresh Token. 
+
+Flow will be completed normally, so Operator provides API invoker with both Access Token and Refresh Token (step 5). 
+ 
+### Refresh Token Flow 
+
+
+```mermaid
+sequenceDiagram
+autonumber
+title Refresh Token Flow - Consent validation
+participant FE as Device App
+participant BE as Invoker<br>(App Backend/Aggregator)
+box Operator
+  participant ExpO as API Exposure Platform<br>@Operator
+  participant Consent as Consent Master<br>@Operator
+end
+
+Note over FE,BE: Use Feature needing<br>Operator Capability  
+
+opt Access to API fails due to Access Token Expired
+  BE->>ExpO: Access Operator CAMARA API <br> Authorization: Bearer {OperatorAccessToken}        
+  ExpO->>ExpO: Decrypt and validate<br>Access Token. It has expired.  
+  ExpO->>BE: 401 Unauthorized<br>{"code": "UNAUTHENTICATED",<br> "message": "Token Expired"}
+end
+
+alt Standard OIDC Refresh Token flow between Invoker and API Exposure Platform
+  BE->>ExpO: POST /token<br>refresh_token={OperatorRefreshToken}
+  ExpO->>ExpO: Validate {OperatorRefreshToken}<br>fetch associated info (scopes)
+  ExpO->>ExpO: Check legal basis of the purpose<br>associated with the token<br> e.g.: contract, legitimate_interest, consent, etc
+  opt If User Consent is required for the legal basis of the purpose  
+      ExpO->>Consent: Check if Consent is granted    
+  end
+  alt If Consent is NOT granted (consent expired/revoked) - The flow fails
+    ExpO->>BE: 400 Bad Request<br>{"error":"invalid_scope"}        
+  else If Consent is Granted or Consent not needed for legal basis   
+    ExpO->>ExpO: Generate<br>new {NewOperatorAccessToken}<br> and {NewOperatorRefreshToken}
+    ExpO->>ExpO: Save new {NewOperatorRefreshToken}
+    ExpO->>BE: 200 OK <br> {NewOperatorAccessToken}<br>{NewOpeatorRefreshToken}
+  end 
+end
+alt Regular Access to API
+  BE->>ExpO: Access Operator CAMARA API <br> Authorization: Bearer {NewOperatorAccessToken}        
+  ExpO->>ExpO: Decrypts NewOperatorAccessToken,<br>grants Access,<br>progresses request to API Backend,<br>gets API response  
+  ExpO->>BE: CAMARA API Response
+  Note over BE,FE: Response
+end
+```
+
+**Key concepts**
+
+- Same considerations that apply for general call flows in [Authorization flows / grant types](#authorization-flows--grant-types) also apply for this flow.
+
+**Flow description**
+
+When a network feature is needed (step 1), the API invoker tries to consume a CAMARA API with an Access Token, but it has already expired, so it gets an error (steps 2-3). 
+
+Upon this error, the API invoker performs standard Refresh Token Grant flow, providing the Operator Refresh Token in order to get a new pair of Operator Refresh Token and Operator Access Token (step 4). 
+
+The Operator validates the refresh token and retrieves the related information, i.e.: scopes/purpose (step 5) and checks the legal basis of the purpose associated with the token (step 6). If the purpose requires the user's consent, the Operator validates whether the consent is granted (step 7). If the consent is NOT granted e.g., the consent has expired or the user has revoked the consent, the Operator will not issue a new token and will return an error (step 8) and the refresh token flow will fail. 
+
+If the user's consent is granted or not required for the applied legal basis, the Operator generates a new pair of Operator Access Token and Aggregator Access Token (step 9), saves the refresh token (step 10), and responds to the API invoker with this information (step 11). 
+
+Then, the API invoker can access normally to the CAMARA API by using the New Operator Access Token (steps 12-14). This part of the flow is just the same as in general call flows.
 
 ## CAMARA API Specification - Authorization and authentication common guidelines
 


AxelNennker:
Why do we need so many words?

OIDC says this about refresh tokens:
```
1.  Offline Access

OpenID Connect defines the following scope value to request offline access:

    offline_access
        OPTIONAL. This scope value requests that an OAuth 2.0 Refresh Token be issued that can be used to obtain an Access Token that grants access to the End-User's UserInfo Endpoint even when the End-User is not present (not logged in). 

When offline access is requested, a prompt parameter value of consent MUST be used unless other conditions for processing the request permitting offline access to the requested resources are in place. The OP MUST always obtain consent to returning a Refresh Token that enables offline access to the requested resources. A previously saved user consent is not always sufficient to grant offline access.

Upon receipt of a scope parameter containing the offline_access value, the Authorization Server:

    MUST ensure that the prompt parameter contains consent unless other conditions for processing the request permitting offline access to the requested resources are in place; unless one or both of these conditions are fulfilled, then it MUST ignore the offline_access request,
    MUST ignore the offline_access request unless the Client is using a response_type value that would result in an Authorization Code being returned,
    MUST explicitly receive or have consent for offline access when the registered application_type is web,
    SHOULD explicitly receive or have consent for offline access when the registered application_type is native.

The use of Refresh Tokens is not exclusive to the offline_access use case. The Authorization Server MAY grant Refresh Tokens in other contexts that are beyond the scope of this specification.
```

A Camara API working group might want to allow or forbid offline access.

A Camara API working group might define a maximum value for the time offline access is valid after the user has given their consent.

Does the IdentityAndConsentManagement group have guide lines for API working groups on how to allow or restrict offline access?

If a user revokes their consent to offline access are our resource servers able to deny access to the API?
Handling of revoked consent MUST be a test case for Camara implementations.

Do we have specific use cases for offline access in current Camara APIs?
If we do not have use cases then we do not need this text, and we might explicitly forbid offline access or at least discourage it.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
 More details about the standard flow can be found in the official IETF specification [The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4).
 
+## Offline Access - Refresh Token
+
+The defined solution uses Refesh Token as a mean to enable CAMARA API access in Offline Scenarios. Online Access means that there is a Telco-based Network Connection, either Mobile Network connection in mobile use cases or Fixed Network connection in fixed use cases.
+
+- Refresh Token may be issued as result of AuthN/AuthZ, if requested by API invoker and allowed by use case. 
+  - [RFC6749 Section 1.5](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5): _A refresh token is a string representing the authorization granted to  the client by the resource owner.  The string is usually opaque to the client. The token denotes an identifier used to retrieve the authorization information. Therefore, the Refresh Token is bound to the authorization of the end user to the Application._
+
+- As indicated by [OIDC Spec](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess ), one of the uses of refresh token is allowing offline access.
```
jogu, line 387: I think this section is conflating things.

The `offline_access` scope is specifically defined in OpenID Connect to mean offline access to 'End-User's UserInfo Endpoint.'

This means when it receives an offline_access request, the authorization server can ask the user if they wish to grant ongoing access to the userinfo.

This same `offline_access` scope should not be extended to have the meaning of requesting equivalent ongoing access to other APIs. If there are other APIs that need to distinguish whether short or long term access is being requested those APIs should define explicit scope values with those meanings.

jpengar: @jogu (and CC @AxelNennker) That's why this usage of the `offline_access` scope is included in the proposed CAMARA OIDC profile (https://github.com/camaraproject/IdentityAndConsentManagement/pull/113), which covers the upcoming GSMA Opengateway technical stream requirement for off-net access scenarios. For example, in scenarios where the device is no longer connected to the operator's network. This information was agreed and specified in the GSMA Opengateway technical stream, and as stated in the PR description, it was agreed to move it to CAMARA.

So this is indeed an extension of `offline_access` not only to the User Info endpoint, but also for Service APIs.

https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md (release v0.1.0), containing CAMARA reached short-term agreements in this WG, also states the usage of `offline_access` for "Off-net" scenarios using refresh_token, as long as there was a connection when the first access_token was requested.

As recommended during today's WG meeting call (14/02), that we first focus on moving the necessary documentation to the CAMARA Identity & Consent WG repository as agreed. And then, if the WG decides so, we are more than happy to discuss this topic for release 0.3.0. focused on enhancements and features for the mid-term.

CC @trehman-gsma @MarkCornall

hdamker: > And then, if the WG decides so, we are more than happy to discuss this topic for release 0.3.0. focused on enhancements and features for the mid-term.

@jpengar I agree that merging the PR and then raise an issue with the needed changes could be the agreed way forward. But if there is a security issue identified and/or the `offline_access` scope can't be extended to other APIs then this MUST be corrected before v0.2.0 can be released.

jpengar: @hdamker What is exactly the security issue you mean?

hdamker: @jpengar nothing concrete as I'm not the expert here, hence the "if". I just wanted to emphasise that we can't postpone critical issues handle as enhancements and postpone to v0.3.0.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
 
 More details about the standard flow can be found in the official IETF specification [The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4).
 
+## Offline Access - Refresh Token
+
+The defined solution uses Refesh Token as a mean to enable CAMARA API access in Offline Scenarios. Online Access means that there is a Telco-based Network Connection, either Mobile Network connection in mobile use cases or Fixed Network connection in fixed use cases.
```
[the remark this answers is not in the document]
palmerabollo: @AxelNennker Could you suggest a clearer way to describe what constitutes an online scenario (the telco is able to identify the connected user) as opposed to offline scenarios (e.g., when the user connects to the Internet via WiFi in a coffee shop)?

documentation/CAMARA-API-access-and-user-consent.md
palmerabollo: This was the agreed solution to **support the off-net scenarios**, e.g., I get an access token when I'm using my mobile connection, then I connect to a Starbucks WiFi and can use & refresh the token. Nobody argued it constitutes a security/privacy issue (let alone critical).

It was also agreed to move the docs from the private GSMA repository, to keep a **single source of truth**, because keeping two versions of the same docs was the worst-case scenario (see #82, https://github.com/camaraproject/IdentityAndConsentManagement/issues/70#issuecomment-1778852651, https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/issues/101, https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/MeetingMinutes/2024/MOM-2024-01-17.md#redundancyoverlap-between-gsma-and-camara-doc-duplicated-discussions, etc).

Knowing that, and that  could we please merge it to get rid of the GSMA version, and then **discuss your alternatives to cover the off-net scenarios**?

documentation/CAMARA-API-access-and-user-consent.md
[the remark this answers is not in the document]
AxelNennker: When the user logs in and gives their consent then the use does not have to be "on-net".

Nit: the telco is in most cases not identifying the "user" but the subscriber.

If using methods .like operator token as defined in section 14 of https://www.gsma.com/newsroom/wp-content/uploads//TS.43-v10.0-Service-Entitlement-Configuration.pdf the device does not have to be "on-net" to identify the subscriber.

I'll try to find a description that fits here.

documentation/CAMARA-API-access-and-user-consent.md
AxelNennker: I think the document should not call "it" offline_access and then talk about refresh tokens. That are different things. (As all know)
I still think that this is wrong and should not be merged.
Something being agreed upon at GSMA does not mean that sections identified now as wrong should be merged in Camara.
I prefer to fix wrongness first and then merge.

AxelNennker:
Could somebody explain why this section is needed at all? I tried to rewrite the chapter but after doing so concluded that this is not needed.

A **normal** OIDC/OAuth2 flow:
Using OIDC or OAuth2 the client, while on-net, gets a short-lived access token and long-lived refresh token for a certain API and scope. The AZ uses some IP-Address-to-MSISDN technology to identify the subscription, which in this scenario is a precondition to getting the access token and refresh token.

In OAuth2, regardless whether the client is on-net or off-net the access token can be used at the RS to use the API.
Now the access token expires. The refresh token can be used to get a new access token for the same scope.
Regardless whether the client is on-net or off-net the new access token can be used at the RS to use the API.

The above is normal for OIDC and OAuth2. The access token is exactly what is needed to access the API. Nothing more, nothing less.

Is there an assumption I am missing? I could imagine that the telco for some APIs requires that API access is on-net. That would be an extension to OAuth2, right? Then not the access token is authorizing the API access but the client has to be on-net for the API access as well. It seems that on-net access to the API is a security feature but not needed to fulfill the function of the API, right?
Otherwise, getting an off-net access token and using it off-net at the RS would not work.

So, my assumption is: For security reasons the RS requires both the access token AND on-net access to the API.
But, there are clients, who can get weaker security and request an off-net capable access token. This is my interpretation why this section we are talking about here exists.

Why refresh tokens play a special role here, I can't imagine.

This sounds like if there is a scope defined by an API and the client requests it, then in fact it is requesting an on-net-scope.
Then to solve the wifi case the client would request scope="openid off-net scope" and would get an access token for off-net access.

Using "offline_access" is confusing for developers because the special scope offline_access creates an access token for the user info endpoint nothing else.

Or am I missing something? Why is this special handling needed?

I fear that if we merge this as a provisional documentation of the current state we never can improve this because then operators who implemented this are going to say "We have a running system with real customers, we cannot change this".

documentation/CAMARA-API-access-and-user-consent.md
palmerabollo: @AxelNennker I disagree for different reasons:
- There was discussion in the GSMA technical stream to find a solution to support off-net scenarios. This was the agreed solution. There was also an **agreement to move that doc from GSMA to CAMARA** to have a single source of truth. This is what this PR does, nothing more.
- **"Wrongness" is an opinion**. There are more solutions in the market implementing similar behaviors. For example, Okta, one of the most popular "identity as a service providers" for devs, does exactly the same
https://developer.okta.com/docs/guides/refresh-tokens/main/#get-a-refresh-token. Microsoft supports the same semantics https://learn.microsoft.com/en-us/entra/identity-platform/scopes-oidc#the-offline_access-scope. Atlassian uses offline_access exactly the same way https://developer.atlassian.com/cloud/oauth/getting-started/refresh-tokens/.  I can't judge whether this is one of those _"other contexts that are beyond the scope of this specification"_ that OIDC leaves the door open to, but it's definitely working well for thousands of devs.

I understand that you might not like it, since every solution has pros&cons (and of course we should keep improving it). But I don't think that's a good reason to block it.

jogu: >  I can't judge whether this is one of those "other contexts that are beyond the scope of this specification" that OIDC leaves the door open to, but it's definitely working well for thousands of devs.

It's helpful to quote the full paragraph so you have the context of those 'other contexts':

> The use of Refresh Tokens is not exclusive to the offline_access use case. The Authorization Server MAY grant Refresh Tokens in other contexts that are beyond the scope of this specification.

This paragraph is making it clear that refresh tokens can be issued even when scope=offline_access is not requested. It is not suggesting that offline_access should be used in other contexts.

For example in OpenData ecosystems, refresh tokens are very commonly used, but I have never seen one that requires the offline_access scope to get a refresh token - it is common for the length of access actually needed by the client application to be indicated to the authorization server though.

As you say, some people do appear to overload offline_access to mean other things. Some people have chosen to not allow applications to indicate which resources they need ongoing access to. However given the wide range of CAMRA APIs I am not sure a decision that a refresh token should only be requested by requesting offline_access is appropriate, and I might go as far as saying that more fine grained permissioning represents current best practice.

jpengar:
>A normal OIDC/OAuth2 flow:
Using OIDC or OAuth2 the client, while on-net, gets a short-lived access token and long-lived refresh token for a certain API and scope. The AZ uses some IP-Address-to-MSISDN technology to identify the subscription, which in this scenario is a precondition to getting the access token and refresh token.
>
>In OAuth2, regardless whether the client is on-net or off-net the access token can be used at the RS to use the API.
Now the access token expires. The refresh token can be used to get a new access token for the same scope.
Regardless whether the client is on-net or off-net the new access token can be used at the RS to use the API.
>
>The above is normal for OIDC and OAuth2. The access token is exactly what is needed to access the API. Nothing more, nothing less.

I fully agree with this. Precisely the use of the refresh_token to get a new access token regardless of whether the client is on-net or off-net is the agreed technical solution to cover the requirements of the Opengateway off-net scenarios.

The point is how to request and/or obtain those long long-lived refresh tokens for a specific CAMARA API (requested scope(s)). 

@jogu @AxelNennker So how exactly do you propose to do that? Are you suggesting to define a specific "off-net"-like scope on each CAMARA API spec? Could you elaborate on that to see if we can find a way to agree on this matter?

If you are suggesting this, it looks to me much more complicated to manage in CAMARA, since it should be included and considered in every CAMARA API subproject. Rather than defining a common way for dozens of APIs and leaving it to the Operator Auth Server to apply the operator's policy to allow or not allow per client_id and/or API if it is allowed to request a refresh token or not. Why should we include this dependency in the API spec? The "openid" scope or "offline_access" scope are not defined at the API level for example... the thing is that OIDC does not define any other endpoint except UserInfo.

If a common way to request a refresh token is defined, how could it be?... using a new specific CAMARA scope, e.g. say "off-net" scope or similar?... But in that case why not directly use "offline_access" which is already defined and is something done by other big players and known by developers. To me, it might not make sense to ask for "offline_access" and "off-net" in the authorization request, for example.

Using "offline_access" to allow clients to request a refresh_token as documented would only need to be considered in the OIDC profiling for CAMARA. Is this really a big deal? What exactly are your security concerns?

AxelNennker:
How about using this text to describe offline access? Does this work with Open Gateway implementations?

### Offline Access

Neither OIDC not OAuth2 define a way for clients to indicate whether they need a refresh_token. 
In OIDC and OAuth2 refresh token issuance is optional in Authorization Code Flow and at the discretion of the AZ.

Camara uses the scope `offline_access` in the authorization request to indicate to the AZ that the client requests a refresh token additionally to the access token for Camara API access.

---
**NOTE**

OIDC defines the scope `offline_access` and it is used to get an access token to be used at the UserInfo endpoint, and also to obtain a request token to refresh this access token.

---

The OIDC rules for `offline_access` as defined in [OIDC Offline Access](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess) apply. 

Copied to here unchanged from the OIDC standard for clarity: 

>  When offline access is requested, a prompt parameter value of consent MUST be used unless other conditions for processing the request permitting offline access to the requested resources are in place. The OP MUST always obtain consent to returning a Refresh Token that enables offline access to the requested resources. A previously saved user consent is not always sufficient to grant offline access.

> Upon receipt of a scope parameter containing the offline_access value, the Authorization Server:

>    MUST ensure that the prompt parameter contains consent unless other conditions for processing the request permitting offline access to the requested resources are in place; unless one or both of these conditions are fulfilled, then it MUST ignore the offline_access request,

>    MUST ignore the offline_access request unless the Client is using a response_type value that would result in an Authorization Code being returned,

>    MUST explicitly receive or have consent for offline access when the registered application_type is web,

>    SHOULD explicitly receive or have consent for offline access when the registered application_type is native.

The OIDC section on [OIDC Using Refresh Tokens](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens) is not changed by this document.

The OIDC security and privacy considerations regarding offline access and refresh tokens apply e.g. on
[Token lifetime](https://openid.net/specs/openid-connect-core-1_0.html#TokenLifetime) and [Offline Access Privacy](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccessPrivacy).

jogu:
> Using "offline_access" to allow clients to request a refresh_token as documented would only need to be considered in the OIDC profiling for CAMARA. Is this really a big deal? What exactly are your security concerns?

As I hinted at above, it's really a fine grained access concern, which gives rise to both security & privacy concerns.

I would say modern best practice would be that clients are very specific about what length of access they need and to which APIs. Whether is it worth addressing that in a generic manner or handling it per API I'm less sure.

jpengar:
> How about using this text to describe offline access? Does this work with Open Gateway implementations?

@AxelNennker Sorry for the late reply, these are tough days. I'm fine with this text, except for the part copied from the OIDC standard... 

> The OIDC rules for offline_access as defined in OIDC Offline Access apply.
>
>Copied here unchanged from the OIDC standard for clarity:
> ...

...mainly because I think these rules, as defined in the standard, do not fit 100% to the CAMARA use case of requesting refresh tokens for Service CAMARA APIs. They are very specific for UserInfo endpoint and Auth Code flow, which is the scenario covered in the standard, and they should be extended in CAMARA OIDC profiling as well for this off-net requirement.

For example:
> MUST ignore the offline_access request unless the client uses a response_type value that would cause an authorization code to be returned,

In the case of CAMARA, Authorization Code is not the only flow to support. For example, offline_access must also be allowed for CIBA flow.

or

> When offline access is requested, a prompt parameter value of consent MUST be used unless there are other conditions for processing the request that allow offline access to the requested resources. The OP MUST always obtain consent to return a refresh token that allows offline access to the requested resources. A previously stored user consent is not always sufficient to grant offline access.

As documented in the PR, there is no requirement on the prompt value to use the "offline_access" scope to request a refresh token to cover Opengateway off-net scenarios to access CAMARA Service APIs.

documentation/CAMARA-API-access-and-user-consent.md
AxelNennker: Is this text acceptable? I think there is no need to mention off-net or on-net. This section is taken from PR https://github.com/camaraproject/IdentityAndConsentManagement/pull/121

I guess the section should not be both in the security and interoperability profile and in CAMARA-API-access-and-user-consent. I think it should be in the "profile" and not in CAMARA-API-access-and-user-consent.

### Offline Access

Neither OIDC not OAuth2 define a way for clients to indicate whether they need a refresh_token. 
In OIDC and OAuth2 refresh token issuance is optional in Authorization Code Flow and at the discretion of the AZ.

Camara uses the scope `offline_access` in the authorization request to indicate to the AZ that the client requests a refresh token additionally to the access token for Camara API access.

---
**NOTE**

OIDC defines the scope `offline_access` and it is used to get an access token to be used at the UserInfo endpoint, and also to obtain a request token to refresh this access token.

---

The OIDC rules for `offline_access` as defined in [OIDC Offline Access](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess) apply. 

Copied to here unchanged from the OIDC standard for clarity: 

>  When offline access is requested, a prompt parameter value of consent MUST be used unless other conditions for processing the request permitting offline access to the requested resources are in place. The OP MUST always obtain consent to returning a Refresh Token that enables offline access to the requested resources. A previously saved user consent is not always sufficient to grant offline access.

> Upon receipt of a scope parameter containing the offline_access value, the Authorization Server:

>    MUST ensure that the prompt parameter contains consent unless other conditions for processing the request permitting offline access to the requested resources are in place; unless one or both of these conditions are fulfilled, then it MUST ignore the offline_access request,

>    MUST ignore the offline_access request unless the Client is using a response_type value that would result in an Authorization Code being returned,

>    MUST explicitly receive or have consent for offline access when the registered application_type is web,

>    SHOULD explicitly receive or have consent for offline access when the registered application_type is native.

The OIDC section on [OIDC Using Refresh Tokens](https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens) is not changed by this document.

The OIDC security and privacy considerations regarding offline access and refresh tokens apply e.g. on
[Token lifetime](https://openid.net/specs/openid-connect-core-1_0.html#TokenLifetime) and [Offline Access Privacy](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccessPrivacy).

AxelNennker (changes requested):
Question: add the proposed new text into the "profile" and not here and close this PR?

jpengar:
> Question: add the proposed new text into the "profile" and not here and close this PR?

I think the final text we end up agreeing definitely needs to be included in the OIDC profile, as discussed earlier in this same PR.

Regarding the refresh_token/offline_access flows being included in CAMARA-API-access-and-user-consent.md in this PR, please don't forget the original goal of this PR and where it came from.

>What this PR does / why we need it:
As discussed and agreed in the CAMARA Identity & Consent Management Working Group (meeting call 17/01/24), and as per CAMARA TSC (Technical Steering Committee) decision, the documentation of the NBI CAMARA API interface, authZ/authN flows, purpose/consent management, etc... will only be placed in CAMARA. CAMARA can be the source of truth for northbound interfaces. GSMA Opengateway will simply adopt what is defined in CAMARA.

>
>This PR is intended to move the GSMA Opengateway [Chapter 04 Offline Access.md](https://github.com/GSMA-Open-Gateway/Open-Gateway-Documents/blob/main/Chapters/Chapter%2004%20Offline%20Access.md) to CAMARA as a last step to fix https://github.com/camaraproject/IdentityAndConsentManagement/issues/82, taking into account PR https://github.com/camaraproject/IdentityAndConsentManagement/pull/114 as well.

Moving this information here was a WG decision after long discussions, including agreement with the GSMA. So not moving this information to CAMARA now, should be a WG decision as well. And If the WG now decides the opposite, at least this information will have to be kept in the "GSMA Opengateway Chapter 04 Offline Access.md" document in the GMSA repository and that document at GSMA will have to be adapted to point to offline access section of CAMARA OIDC profile when available.

hdamker:
> information will have to be kept in the "GSMA Opengateway Chapter 04 Offline Access.md" document in the GMSA repository and that document at GSMA will have to be adapted to point to offline access section of CAMARA OIDC profile when available.

If there is a consensus possible based on the text proposed by Axel (multiple times above) than this is maybe the way forward: keep the text for now in GSMA and add the agreed, adapted version of it to the OIDC profile. 

At least it wouldn't make sense to have too different statements for the topic of refresh token and their use in offline scenarios in two different documents of the working group (even not for a short time).

AxelNennker:
In CAMARA-Security-Interoperability.md the Offline Access section is a subsection of Authorization Code flow. So, only applies to that flow.

CIBA has its own rules for refresh tokens at least in PUSH mode e.g.:
> In case a Refresh Token is sent to the Client, the hash value of it MUST also be added to the ID token using the urn:openid:params:jwt:claim:rt_hash claim.

I think the proposed text is good for authorization code flow.
Have to think about ciba and refresh tokens.
But I would rather have the proposed text for authorziation code flow then the GSMA text.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
+
+**Flow description**
+
+When a network feature is needed (step 1), the API invoker tries to consume a CAMARA API with an Access Token, but it has already expired, so it gets an error (steps 2-3). 
+
+Upon this error, the API invoker performs standard Refresh Token Grant flow, providing the Operator Refresh Token in order to get a new pair of Operator Refresh Token and Operator Access Token (step 4). 
+
+The Operator validates the refresh token and retrieves the related information, i.e.: scopes/purpose (step 5) and checks the legal basis of the purpose associated with the token (step 6). If the purpose requires the user's consent, the Operator validates whether the consent is granted (step 7). If the consent is NOT granted e.g., the consent has expired or the user has revoked the consent, the Operator will not issue a new token and will return an error (step 8) and the refresh token flow will fail. 
+
+If the user's consent is granted or not required for the applied legal basis, the Operator generates a new pair of Operator Access Token and Aggregator Access Token (step 9), saves the refresh token (step 10), and responds to the API invoker with this information (step 11).
```
AxelNennker, line 556: This text can't be added because it adds `Aggregator Access Token` and `Operator Access Token` which are unknown concepts in Camara.

AxelNennker (changes requested):

documentation/CAMARA-API-access-and-user-consent.md
jpengar: @AxelNennker My bad, one of the things I did before moving the documentation was to remove any concept unknown to CAMARA like the aggregation model of GSMA Opengateway. Good catch, thanks. I will review the PR to correct it.

jpengar: Typo fixed in https://github.com/camaraproject/IdentityAndConsentManagement/pull/123/commits/354d61853ede084cc57d42e14e4affea02ad5078.

AxelNennker (changes requested):
I think there is no such thing as an `Operator Access Token`. We only have access tokens. I think that applies to the whole GSMA text. I guess every occurrence of "operator" is a candidate for a review.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
+
+The API invoker (which could potentially be the application backend, an aggregator, etc.) instructs the application frontend in the device to initiate the OIDC authorization code flow with the operator as described in the [Authorization code flow (Frontend flow)](#authorization-code-flow-frontend-flow).   
+
+The API invoker includes `offline_access` scope to signal that a Refresh Token is expected (step 1). 
+
+Then same flow depicted in [Authorization code flow (Frontend flow)](#authorization-code-flow-frontend-flow) occurs (steps 2-8). 
+
+> **NOTE: Flow diagram is simplified, full details offered in general call flows** 
+
+When Operator is going to issue the Access Token, it also issues a Refresh Token (OperatorRefreshToken), as `offline_access` scope was requested.
```
AxelNennker, line 441:
```suggestion
When the Authorization Server is going to issue the Access Token, it also issues a Refresh Token, as `offline_access` scope was requested. 
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
+```
+
+**Key concepts**
+
+- Same considerations that apply for general call flows in [Authorization flows / grant types](#authorization-flows--grant-types) also apply for this flow. 
+- Refresh Token issuance is achieved by requesting `offline_access` scope 
+ 
+**Flow description**
+
+The API invoker (which could potentially be the application backend, an aggregator, etc.) requests a 3-legged access token to the operator API exposure platform. The process follows the OpenID Connect Client-Initiated Backchannel Authentication (CIBA) flow as described in [CIBA flow (Backend flow)](#ciba-flow-backend-flow).
```
AxelNennker, line 487:
```suggestion
The API invoker (which could potentially be the application backend, an aggregator, etc.) requests a 3-legged access token to the API exposure platform. The process follows the OpenID Connect Client-Initiated Backchannel Authentication (CIBA) flow as described in [CIBA flow (Backend flow)](#ciba-flow-backend-flow).  
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
+
+**Key concepts**
+
+- Same considerations that apply for general call flows in [Authorization flows / grant types](#authorization-flows--grant-types) also apply for this flow.
+
+**Flow description**
+
+When a network feature is needed (step 1), the API invoker tries to consume a CAMARA API with an Access Token, but it has already expired, so it gets an error (steps 2-3). 
+
+Upon this error, the API invoker performs standard Refresh Token Grant flow, providing the Operator Refresh Token in order to get a new pair of Operator Refresh Token and Operator Access Token (step 4).
```
AxelNennker, line 552:
```suggestion
Upon this error, the API invoker performs standard Refresh Token Grant flow, providing the Refresh Token in order to get a new pair of Refresh Token and Access Token (step 4). 
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
+
+- Same considerations that apply for general call flows in [Authorization flows / grant types](#authorization-flows--grant-types) also apply for this flow.
+
+**Flow description**
+
+When a network feature is needed (step 1), the API invoker tries to consume a CAMARA API with an Access Token, but it has already expired, so it gets an error (steps 2-3). 
+
+Upon this error, the API invoker performs standard Refresh Token Grant flow, providing the Operator Refresh Token in order to get a new pair of Operator Refresh Token and Operator Access Token (step 4). 
+
+The Operator validates the refresh token and retrieves the related information, i.e.: scopes/purpose (step 5) and checks the legal basis of the purpose associated with the token (step 6). If the purpose requires the user's consent, the Operator validates whether the consent is granted (step 7). If the consent is NOT granted e.g., the consent has expired or the user has revoked the consent, the Operator will not issue a new token and will return an error (step 8) and the refresh token flow will fail.
```
AxelNennker, line 554:
```suggestion
The Authorization Server validates the refresh token and retrieves the related information, i.e.: scopes/purpose (step 5) and checks the legal basis of the purpose associated with the token (step 6). If the purpose requires the user's consent, the Operator validates whether the consent is granted (step 7). If the consent is NOT granted e.g., the consent has expired or the user has revoked the consent, the Authorization Server will not issue a new token and will return an error (step 8) and the refresh token flow will fail. 
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
+
+**Flow description**
+
+When a network feature is needed (step 1), the API invoker tries to consume a CAMARA API with an Access Token, but it has already expired, so it gets an error (steps 2-3). 
+
+Upon this error, the API invoker performs standard Refresh Token Grant flow, providing the Operator Refresh Token in order to get a new pair of Operator Refresh Token and Operator Access Token (step 4). 
+
+The Operator validates the refresh token and retrieves the related information, i.e.: scopes/purpose (step 5) and checks the legal basis of the purpose associated with the token (step 6). If the purpose requires the user's consent, the Operator validates whether the consent is granted (step 7). If the consent is NOT granted e.g., the consent has expired or the user has revoked the consent, the Operator will not issue a new token and will return an error (step 8) and the refresh token flow will fail. 
+
+If the user's consent is granted or not required for the applied legal basis, the Operator generates a new pair of Operator Access Token and Operator Refresh Token (step 9), saves the refresh token (step 10), and responds to the API invoker with this information (step 11).
```
AxelNennker, line 556:
```suggestion
If the user's consent is granted or not required for the applied legal basis (purpose), the Authorization Server generates a new pair of Access Token and Refresh Token (step 9), saves the refresh token (step 10), and responds to the API invoker with this information (step 11). 
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -373,6 +377,186 @@ The Client Credentials grant type is used to obtain a 2-legged access_token that
...
+```
+
+**Key concepts**
+
+- Same considerations that apply for general call flows in [Authorization flows / grant types](#authorization-flows--grant-types) also apply for this flow.
+-  Refresh Token issuance is achieved by requesting `offline_access` scope. 
+
+**Flow description** 
+
+The API invoker (which could potentially be the application backend, an aggregator, etc.) instructs the application frontend in the device to initiate the OIDC authorization code flow with the operator as described in the [Authorization code flow (Frontend flow)](#authorization-code-flow-frontend-flow).
```
AxelNennker, line 433:
```suggestion
The API invoker (which could potentially be the application backend, an aggregator, etc.) instructs the application frontend in the device to initiate the OIDC authorization code flow with the Authorization Server as described in the [Authorization code flow (Frontend flow)](#authorization-code-flow-frontend-flow).   
```

AxelNennker (changes requested):

jpengar:
> I think there is no such thing as an `Operator Access Token`. We only have access tokens. I think that applies to the whole GSMA text. I guess every occurrence of "operator" is a candidate for a review.

@AxelNennker The operator access token is actually the access token (access tokens are issued by the operator). It is just a matter of naming, and the PR text is aligned with the rest of the document content. If we want to revisit this naming later and change it throughout the document. That's fine with me. But I don't see that as a blocker for this PR now.

AxelNennker:
I still hope that we can have the offline access / refresh token text in https://github.com/camaraproject/IdentityAndConsentManagement/pull/121 and not here. I think the GSMA text is too operator specific. I suggested some changes but even with them applied this GSMA text appears to be to not-necessary 
I think the text I added this morning to #121  covers refresh token usage in authorization code flow and CIBA - there really no difference. I am going to review the refresh token _issuance_ text again. That is part of authorization code flow section and there text for CIBA is probably written by me today.

AxelNennker:
Closing this as agreed during the March 13th ICM meeting.

PR #158, opened 2024-05-22
Fix statement about "missing sub claim" in case there is no id token

### LINKED ISSUE #156: Fix statement about "missing sub claim" in case there is no id token
**Problem description**
The document CAMARA-Security-Interoperability.md contains a statement which is not correct. The current statement is:
_Clients SHOULD follow the OIDC standard and SHOULD include openid in the list of requested scopes. Without id token there is no sub field and the privacy features of OIDC are severely crippled. Globally unique identifiers, like the MSISDN, should be avoided for privacy reasons._

It is correct that without requesting openid scope, there is no id token generated by the authZ server.
However it is NOT correct that the access token does not contain a sub claim in case there is not id token.

The sub claim of the access token is filled with an identifier of the subject / resource owner in case of 3-legged flow and with the client_id or application id in case of 2-legged flow.

It is necessary to give an argumentation on why an id token (as requested by openid scope) should be used. 
One argument could be that the sub claim in the id token contains an obfuscated identifier instead of a globally unique identifier for the subject like MSISDN. This identifer can only be interpreted by the authZ server.

**Expected action**
1) Remove wrong statement: _Without id token there is no sub field and the privacy features of OIDC are severely crippled._ 
with a correct statement about the sub field.
2) Give an argumentation on why an id token is proposed.
3) Clarify the content of the sub claim of an access token, e.g.: 
The sub field of the access token is meant to identify the subject / resource owner in case of 3-legged flow and the client_id or application id in case of 2-legged flow. Globally unique identifiers, like the MSISDN, should be avoided for privacy reasons._

**Additional context**

AxelNennker:
#### What type of PR is this?

* correction

Fixes #156

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -208,7 +208,7 @@ If "openid" is missing in the scope value but a claim that is [standardized in O
 Clients SHOULD follow the OIDC and CIBA standard and SHOULD include `openid` in the list of requested scopes.
 The [id token](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) contains the `sub` field which is the identifier of the subject of the [OIDC authorization code](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) request respectively the [CIBA authentication request](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_request). 
 
-The Camary authorization server creates the `sub` value. Globally unique identifiers, like the MSISDN, should be avoided for privacy reasons.
+The Camara authorization server creates the `sub` value. Globally unique identifiers, like the MSISDN, should be avoided for privacy reasons.
 
 ## Purpose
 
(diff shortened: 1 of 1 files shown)


documentation/CAMARA-Security-Interoperability.md
```diff
@@ -205,9 +205,10 @@ Please refer to [Authentication Error Response](https://openid.net/specs/openid-
...
 
-Clients SHOULD follow the OIDC standard and SHOULD include `openid` in the list of requested scopes.
-Without id token there is no `sub` field and the privacy features of OIDC are severely crippled. 
-Globally unique identifiers, like the MSISDN, should be avoided for privacy reasons.
+Clients SHOULD follow the OIDC and CIBA standard and SHOULD include `openid` in the list of requested scopes.
+The [id token](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) contains the `sub` field which is the identifier of the subject of the [OIDC authorization code](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) request respectively the [CIBA authentication request](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_request). 
+
+The Camary authorization server creates the `sub` value. Globally unique identifiers, like the MSISDN, should be avoided for privacy reasons.
```
Elisabeth-Ericsson, line 211:
```suggestion
The Camara authorization server creates the `sub` value. Globally unique identifiers, like the MSISDN, should be avoided for privacy reasons.
```
Agree to the new wording, but pls fix the small type'o

jpengar (commented):
/LGTM

Elisabeth-Ericsson (approved):
/LGTM

AxelNennker:
@jpengar please "/LGTM" instead of "LGTM"

jpengar:
> @jpengar please "/LGTM" instead of "LGTM"

And why is that? I'm just curious, because I approved the PR anyway... What is the difference?

PR #169, opened 2024-05-30
Delete code directory

### LINKED ISSUE #166: code folder to be removed from ICM working group ?
**Problem description**
The ICM repo (as all other repos) was created from a template that contains folders for Code/API-definitions etc.
A question was raised if the ICM group is going to create APIs or not ? The fact that the folders are there seem to indicate it is the case, and could have people waiting for ICM to publish APIs.

If is is clear that (at least for now) no APIs will be created in this working group, the ask is to remove the code folder from the repo.
NOTE: I would actually expect that if there are APIs to be defined, there should be a dedicated API Sub Project for that such that their release management is separate from the ICM documentation release anyway.

It may need to be checked that ICM documentation clearly indicates the expectation that operators are to use the APIs / methods supported by their implementation of the ICM related aspects.

**Expected action**
If agreed, remove the code folder from the repo.
Check ICM documentation on expectation for usage of ICM implementation APIs

**Additional context**

AxelNennker:
#### What type of PR is this?

* cleanup

As Tanja explains in #166 ICM does not produce any API.

Fixes #166

### DIFF
--- a/code/API_definitions/README.MD
+++ b/code/API_definitions/README.MD
@@ -1,1 +1,0 @@
-Here you can add your definitions and delete this README.MD file


jpengar (approved):
LGTM. I'm fine with removing the code folder for now.

Issue #98, opened 2023-12-19
Terms used in User Identity section in CAMARA-API-access-and-user-consent.md


bhjelm:
**Problem description**
Update the terms used in User Identity section of [CAMARA-API-access-and-user-consent.md](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#user-identity.)

**Expected action**
The current text isn't clear and mixes the use of (user) "identity," "identification," "identifier" that should be clearly specified in this section (as well as added to the Glossary of Terms section). 

**Additional context**
As suggestion could be to use Section 3.1, [3GPP TR 22.904](https://portal.3gpp.org/desktopmodules/Specifications/SpecificationDetails.aspx?specificationId=3417), as a guide (for User/User Identity/User Identifier) as well as Figure 1, Section 4.2, that depicts the relationship between user, identities, identifiers and attributes. The terminology used also aligns with ISO/IEC 24760 (IT Security and Privacy — A framework for identity management).

jpengar:
Moving comment from @murthygorty in [#93](https://github.com/camaraproject/IdentityAndConsentManagement/pull/93#pullrequestreview-1789857678) here, which is a related topic:

>In terms and concepts, can you consider changing "User" term to "Subscriber" please? If approved and changed, we will have
>
>End User -> human participant using the application (can be a parent or alternate delegate of the carrier-resource)
>Subscriber -> subscriber of the telco operator, owner of the carrier-resource at hand. (eg., parent)

jpengar:
As stated in https://github.com/camaraproject/IdentityAndConsentManagement/issues/88, we are in the process of finalizing a first release v0.1 of the documentation with the existing agreements. But apart from that, from Telefónica side, we are more than happy if you guys contribute with a PR to the documentation once the v0.1 release is closed, to add the suggested improvement for everyone to review.

PR #173, opened 2024-06-04
Proposed text on network-based authentication

AxelNennker:
There is no OAuth sub in [OAuth2](https://datatracker.ietf.org/doc/html/rfc6749).
There is an id_token.sub in OIDC.

Network authentication identifies the subscription which is primarily defined by the IMSI.
If there is a one-to-one relationship between between IMSI and MSISDN then network authentication also identifies the MSISDN. If one subscriber has multiple SIMs with the same MSISDN then things might go wrong for the Camara API, because the user might want some QoD for the device they are currently using but not for the other with the same.

#### What type of PR is this?

* correction

- replace OAuth sub
- Describe what network authentication does 
- Clarify that globally unique and long-term identifiers like MSISDN must not be handed to the API consumer in id_token.sub. Also associate `sub` with the access token.

### DIFF
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -140,7 +140,7 @@ As per the standard authorization code flow, the device application is redirecte
 
 The API exposure platform receives the request from the device application (Step 3) and does the following:
 
-- Use network based authentication mechanism to obtain the subscription identifier, i.e.: MSISDN or IMSI. Set the id_token sub to some unique user ID and associate the sub with the access token. The id_token sub SHOULD not reveal information to the API consumer that they not already know, e.g. using the MSISDN as a sub might violate privacy. (Step 4).
+- Use network based authentication mechanism to obtain the subscription identifier,e.g.: phone number or IMSI. Set the id_token sub to some unique user ID and associate the sub with the access token. The id_token sub SHOULD NOT reveal information to the API consumer that they not already know, e.g. using the MSISDN as a sub might violate privacy. (Step 4).
 
 - Check if user consent is required, which depends on the legal basis associated with the purpose ("legitimate interest", "contract", "consent", etc). If necessary, it will check in the operator's consent master whether user consent has already been given for this identifier, the application client_id and the requested purpose (Steps 5-6).
 
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -140,7 +140,7 @@ As per the standard authorization code flow, the device application is redirecte
 
 The API exposure platform receives the request from the device application (Step 3) and does the following:
 
-- Use network based authentication mechanism to obtain the user identifier, i.e.: MSISDN. Set the OAuth sub to the unique user ID in the operator (Step 4).
+- Use network based authentication mechanism to obtain the subscription identifier, i.e.: MSISDN or IMSI. Set the id_token sub to some unique user ID and associate the sub with the access token. The id_token sub SHOULD not reveal information to the API consumer that they not already know, e.g. using the MSISDN as a sub might violate privacy. (Step 4).
 
 - Check if user consent is required, which depends on the legal basis associated with the purpose ("legitimate interest", "contract", "consent", etc). If necessary, it will check in the operator's consent master whether user consent has already been given for this identifier, the application client_id and the requested purpose (Steps 5-6).
 


jpengar:
> In the current version of https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md everything hinges on network-based authentication.
Which is not enough for user consent because network-based "authentication" identifies the subscription not the user.
Maybe if "prompt=none" in combination with a some "purpose", is network-based authentication enough to decide whether user consent is "needed" or not. Otherwise the user must authenticate using some login-credentials and give consent.

Please have a look at what the document defines as `User` and as `End User`.

* _End User: the human participant who uses the application from a consumption device._
* _User: the client/subscriber of the telco operator, identified by a unique user identifier (e.g. subject identifier sub in OpenID Connect terminology). The user is the resource owner. Usually the user corresponds to the end user, but this is not always the case. For example, a parent may be the user of a mobile subscription for their children._

So please note that when the document refers to the user, it means the Telco Operator subscriber.

AxelNennker:
Clarified the PR description.

AxelNennker:
> > In the current version of https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md everything hinges on network-based authentication.
> > Which is not enough for user consent because network-based "authentication" identifies the subscription not the user.
> > Maybe if "prompt=none" in combination with a some "purpose", is network-based authentication enough to decide whether user consent is "needed" or not. Otherwise the user must authenticate using some login-credentials and give consent.
> 
> Please have a look at what the document defines as `User` and as `End User`.
> 
>     * _End User: the human participant who uses the application from a consumption device._
> 
>     * _User: the client/subscriber of the telco operator, identified by a unique user identifier (e.g. subject identifier sub in OpenID Connect terminology). The user is the resource owner. Usually the user corresponds to the end user, but this is not always the case. For example, a parent may be the user of a mobile subscription for their children._
> 
> 
> So please note that when the document refers to the user, it means the Telco Operator subscriber.



> > In the current version of https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md everything hinges on network-based authentication.
> > Which is not enough for user consent because network-based "authentication" identifies the subscription not the user.
> > Maybe if "prompt=none" in combination with a some "purpose", is network-based authentication enough to decide whether user consent is "needed" or not. Otherwise the user must authenticate using some login-credentials and give consent.
> 
> Please have a look at what the document defines as `User` and as `End User`.
> 
>     * _End User: the human participant who uses the application from a consumption device._
> 
>     * _User: the client/subscriber of the telco operator, identified by a unique user identifier (e.g. subject identifier sub in OpenID Connect terminology). The user is the resource owner. Usually the user corresponds to the end user, but this is not always the case. For example, a parent may be the user of a mobile subscription for their children._
> 
> 
> So please note that when the document refers to the user, it means the Telco Operator subscriber.

I am not happy with the definition of "user" in documentation/CAMARA-API-access-and-user-consent.md because I find it confusing and contrary to what is usually defined to be a user by non-telco readers, but that is not part of this PR.
Readers who are API consumers might be surprised by this definition.

I say that mixing user and subscriber is a typical telco. Camara API Design guidelines tell us to avoid t[elco terminology](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#25-reduce-telco-specific-terminology-in-api-definitions). I, personally, mixing subscriber and user falls under that guideline.
Again not part of this PR.

sebdewet:
> * _End User: the human participant who uses the application from a consumption device._
> * _User: the client/subscriber of the telco operator, identified by a unique user identifier (e.g. subject identifier sub in OpenID Connect terminology). The user is the resource owner. Usually the user corresponds to the end user, but this is not always the case. For example, a parent may be the user of a mobile subscription for their children._

I would have just said for the End User :
the human participant who uses the device. (can be through an application, a browser etc.)

And I would prefer resource owner for User.

AxelNennker:
> > * _End User: the human participant who uses the application from a consumption device._
> > * _User: the client/subscriber of the telco operator, identified by a unique user identifier (e.g. subject identifier sub in OpenID Connect terminology). The user is the resource owner. Usually the user corresponds to the end user, but this is not always the case. For example, a parent may be the user of a mobile subscription for their children._
> 
> I would have just said for the End User : the human participant who uses the device. (can be through an application, a browser etc.)
> 
> And I would prefer resource owner for User.

@sebdewet please create a PR for the glossary if you think that should be a change.
Are you OK with my new text in this PR?

> - Use network based authentication mechanism to obtain the subscription identifier, i.e.: MSISDN or IMSI. Set the id_token sub to some unique user ID and associate the sub with the access token. The id_token sub SHOULD not reveal information to the API consumer that they not already know, e.g. using the MSISDN as a sub might violate privacy. (Step 4).

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -140,7 +140,7 @@ As per the standard authorization code flow, the device application is redirecte
 
 The API exposure platform receives the request from the device application (Step 3) and does the following:
 
-- Use network based authentication mechanism to obtain the user identifier, i.e.: MSISDN. Set the OAuth sub to the unique user ID in the operator (Step 4).
+- Use network based authentication mechanism to obtain the subscription identifier, i.e.: MSISDN or IMSI. Set the id_token sub to some unique user ID and associate the sub with the access token. The id_token sub SHOULD not reveal information to the API consumer that they not already know, e.g. using the MSISDN as a sub might violate privacy. (Step 4).
```
jpengar, line 143:
```suggestion
- Use network based authentication mechanism to obtain the user identifier,e.g.: phone number or IMSI. Set the id_token sub to some unique user ID and associate the sub with the access token. The id_token sub SHOULD not reveal information to the API consumer that they not already know, e.g. using the MSISDN as a sub might violate privacy. (Step 4).
```

AxelNennker: That "user" is based on the definition in the glossary. I think we should write the correct term here which is subscriber.

jpengar: Well, I think the text should be consistent with the glossary terms. Changing the glossary terms is another discussion.

AxelNennker: I think using the term subscription is the correct one and does not require a glossary change.
The network authentication never authenticates the end-user and using "user" here is confusing.

AxelNennker:
I think that "user" is not as specific as it can be here. Only the "unusall" definition of user in the glossary make the use of the term "user" workable. But why avoid the correct term "subscription"?

documentation/CAMARA-API-access-and-user-consent.md
jpengar: Ok, updated suggestion. I keep "subscription" in this sentence.

jpengar:
> I think that "user" is not as specific as it can be here. Only the "unusall" definition of user in the glossary make the use of the term "user" workable. But why avoid the correct term "subscription"?

If we commit my suggestion [above](https://github.com/camaraproject/IdentityAndConsentManagement/pull/173#discussion_r1627853439) (keeping the term "subscription" in that sentence), I'm fine with merging the PR. And then we can discuss the correct term to use along the document in another issue/PR as suggested [here](https://github.com/camaraproject/IdentityAndConsentManagement/pull/173#issuecomment-2149032343) by @AxelNennker. That change would require reviewing all text along the document and we would be better off doing that in a separate PR if eventually needed. This is actually related to https://github.com/camaraproject/IdentityAndConsentManagement/issues/98.

In Axel's absence I could commit my suggestion and merge this PR I think, if you @sebdewet are okay with it and with what I mentioned above.

jpengar:
@sebdewet Your review here is needed to merge the PR as agreed in [the June 19 WG meeting call](https://wiki.camaraproject.org/pages/viewpage.action?pageId=31227941). Could you please take a look?

jpengar (approved):
LGTM

jpengar:
I'm merging this PR as agreed in [the June 19 WG meeting call](https://wiki.camaraproject.org/pages/viewpage.action?pageId=31227941), since there have been no further comments or feedback.

sebdewet:
LGTM

PR #180, opened 2024-07-12
Fix returned error in authentication code flow

garciasolero:
#### What type of PR is this?

* correction

According to OAuth 2.0 ([RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1)), if the resource owner denies the request, the response should redirect the user-agent to the client redirection URI by adding the `error` parameter with `access_denied` value.

> If the resource owner denies the access request or if the request fails for reasons other than a missing or invalid redirection URI, the authorization server informs the client by adding the following parameters to the query component of the redirection URI using the"application/x-www-form-urlencoded" format [...]


jpengar (approved):
LGTM. Good catch! That was a doc typo.

sebdewet (approved):
LGTM !

AxelNennker (approved):

PR #182, opened 2024-07-26
Improve Access and User Consent document

chrishowell:
#### What type of PR is this?
documentation

#### What this PR does / why we need it:
Defines terms and keeps them consistent throughout, improves clarity in English writing.

#### Which issue(s) this PR fixes:

https://github.com/camaraproject/IdentityAndConsentManagement/issues/98

#### Changelog input

```
 release-note access-and-user-consent documentation update

```

#### Additional documentation 

```
docs

```

PR #183, opened 2024-07-30
Editorial changes to CAMARA-API-access-and-user-consent.md

AxelNennker:
#### What type of PR is this?

* documentation

This PR takes the dead-simple, minor improvements from https://github.com/camaraproject/IdentityAndConsentManagement/pull/182 into their own PR.

Makes reviewing https://github.com/camaraproject/IdentityAndConsentManagement/pull/182 simpler.
This is a very small PR, which I would include into rc.2 because it is (some) improvement and not much work and we are doing rc.2 anyway.


jpengar (approved):
LGTM

PR #210, opened 2024-10-07
error_response_missing_openid_scope

### LINKED ISSUE #209: OIDC authorization code flow error response
**Problem description**
CAMARA-Security-Interoperability.md states that if the openid scope is missing then a 400 BAD request should be returned.
A HTTP response of 400 must not be used.

**Expected behavior**
Errors should be returned as specified in the OIDC standard.
That is, if the redirect_url is valid, then the error is conveyed through the redirect_url as defined in [3.1.2.6.  Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError)

AxelNennker:
#### What type of PR is this?

* correction

Error responses should be returned through the redirect_url if the redirect_url is valid.
Only if the redirect_url is not valid are HTTP response codes used.

Fixes #209

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -195,15 +195,18 @@ Therefore scopes should be available to API implementations.
 
 ## Missing "openid" scope
 
-[OIDC Core Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) states the following about the value of scope.
+[OIDC Connect Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) states the following about the value of scope.
 > REQUIRED. OpenID Connect requests MUST contain the openid scope value. If the openid scope value is not present, the behavior is entirely unspecified. Other scope values MAY be present. ...
 
 This document defines that the authentication server SHOULD not return an id token if `openid` is missing in the scope parameter.
 
 This document defines the following error handling for a missing "openid" value in scope. 
-Please refer to [Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).
 
+For OIDC please refer to [OIDC Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).
 If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an error `invalid_request`.
+
+For CIBA please refer to [CIBA Authentication Error Response](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_error_response).
+If "openid" is missing in the scope value but a claim that is standardized in OIDC, then the Authorization Server returns an error `invalid_request` with an HTTP status code of 400 BAD_REQUEST.
 
 Clients SHOULD follow the OIDC and CIBA standard and SHOULD include `openid` in the list of requested scopes.
 The [id token](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) contains the `sub` field which is the identifier of the subject of the [OIDC authorization code](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) request respectively the [CIBA authentication request](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_request). 
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -203,7 +203,7 @@ This document defines that the authentication server SHOULD not return an id tok
 This document defines the following error handling for a missing "openid" value in scope. 
 Please refer to [Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).
 
-If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an HTTP response code of 400 (Bad Request) and an error invalid_request.
+If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an error `invalid_request`.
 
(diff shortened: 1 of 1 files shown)


sebdewet:
Hi, 
not feasable in ciba when we won't have a redirect_uri in poll mode.
When a redirect_uri is specified, you want to return the errors in a query string using an HTTP redirect 302 status code ?

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -203,7 +203,7 @@ This document defines that the authentication server SHOULD not return an id tok
 This document defines the following error handling for a missing "openid" value in scope. 
 Please refer to [Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).
 
-If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an HTTP response code of 400 (Bad Request) and an error invalid_request.
+If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an error `invalid_request`.
```
jpengar, line 206: @AxelNennker I understand that you are changing the phrase for a more generic one that only refers to `invalid_request` error, but does not specify whether it should be an HTTP error or not, because in the case of auth code for example, it would be provided in the redirect URL. Right?

AxelNennker: In OIDC authorization code flow, as long as the redirect_uri, is valid the error is returned via the redirect_uri.
```
HTTP/1.1 302 Found
Location: https://client.example.org/cb?
    error=invalid_request
    &state=af0ifjsldkj
```

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -195,15 +195,18 @@ Therefore scopes should be available to API implementations.
 
 ## Missing "openid" scope
 
-[OIDC Core Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) states the following about the value of scope.
+[OIDC Connect Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) states the following about the value of scope.
```
jpengar, line 198: `OIDC` already includes Connect... and why do you replace "Core" with "Connect"? Another option could be:
```suggestion
[OpenID Connect Core Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) states the following about the value of scope.
```
But I see it ok as it was `OIDC Core Authentication Request`

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -195,15 +195,18 @@ Therefore scopes should be available to API implementations.
...
 > REQUIRED. OpenID Connect requests MUST contain the openid scope value. If the openid scope value is not present, the behavior is entirely unspecified. Other scope values MAY be present. ...
 
 This document defines that the authentication server SHOULD not return an id token if `openid` is missing in the scope parameter.
 
 This document defines the following error handling for a missing "openid" value in scope. 
-Please refer to [Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).
 
-If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an HTTP response code of 400 (Bad Request) and an error invalid_request.
+For OIDC please refer to [OIDC Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).
```
jpengar, line 205:
```suggestion
For the Authorization Code Flow please refer to [OIDC Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError).
```
We can say  Authorization Code Flow instead of OIDC. CIBA is also OpenId Connect. And the link points to section `3.1.2.6.  Authentication Error Response` of the OIDC Core standard which is under section `3.1.  Authentication using the Authorization Code Flow`.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -195,15 +195,18 @@ Therefore scopes should be available to API implementations.
...
+If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an error `invalid_request`.
+
+For CIBA please refer to [CIBA Authentication Error Response](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_error_response).
+If "openid" is missing in the scope value but a claim that is standardized in OIDC, then the Authorization Server returns an error `invalid_request` with an HTTP status code of 400 BAD_REQUEST.
```
jpengar, line 209:
```suggestion
If "openid" is missing in the scope value but a claim that is [standardized in OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Claims) is requested, then the Authorization Server returns an `invalid_request` error code.

For CIBA please refer to [CIBA Authentication Error Response](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_error_response).
If "openid" is missing in the scope value but a claim that is standardized in OIDC, then the Authorization Server returns an HTTP 400 Bad request response with an `invalid_request` error code.
```

documentation/CAMARA-Security-Interoperability.md
AxelNennker: I think CIBA is not OIDC (OpenId Connect). They are separate standards published by the OIDF.

jpengar: Yes, of course, it is indeed a separate standard. What I mean is that it is called "**OpenID Connect** Client-Initiated Backchannel Authentication Flow - Core 1.0", so it looks better (or at least helpful) to me to explicitly mention the Authorization Code Flow instead of OIDC in this case, to avoid misunderstandings.

jpengar: But in any case, I don't think it's a big deal, so I've given my approval.

jpengar (approved):

AxelNennker:
I think the text is now clear about CIBA and OIDC authoritzation code flow, thus adressing @sebdewet 's comment.
Merging this now. And closing.

PR #220, opened 2024-10-18
Add response codes for error scenarios

### LINKED ISSUE #211: New sections with error scenarios
**Problem description**

In several integrations we have found that, with the existing profile, application developers still have difficulty identifying which errors they will receive in their apps depending on the failure scenario. We know that many scenarios are already conveniently explained in the documentation of the different standards, but we feel that it would be beneficial to have a single place where errors can be consulted at a glance, without having to navigate through different specifications. This would reduce misunderstandings and potential interoperability issues between operators.

**Expected action**

We believe it would be convenient to have sections in the profile (or in a separate document) that compile the possible errors for each endpoint for the different flows. For instance:

**CIBA Token endpoint:**

| Status Code           | Error Code         | Scenario                                                                                                                             |       
|--------------------|------------------| --------------------------------------------------------------------------------------------|                                                                                                                                                                                                                                                                                                                 
| 400 - Bad Request | `access_denied`  | The user denied the authorization request (for example, there is no consent from the user).|
[...]


**Additional context**
<!-- Add any other context of the documentation issue e.g. reference documents. -->

garciasolero:
#### What type of PR is this?

* documentation

We know that many scenarios are already conveniently explained in the documentation of various standards, but a new section compiling errors and their scenarios from several specifications would reduce misunderstandings and potential interoperability issues between operators.

Fixes #211


jpengar (dismissed):
LGTM

eric-murray (changes requested):
Following discussion at the TSC meeting last week, we need to be clear which parts of our documents are normative and which are informative.

So is this new section intended to be normative? - i.e. it will apply, even if it turns out that OIDC says something different compared to this section.

Or is it informative? So a summary of how we think OIDC errors apply to the CAMARA scenario, but OIDC remains normative?

The use of "MUST" in the first sentence implies normative, in which case, that status should be added in brackets to the section header e.g. "Appendix A (Normative) : Error Responses"

jpengar:
As discussed [in the last WG call](https://lf-camaraproject.atlassian.net/l/cp/8xpkddMk), this proposal is to make the definitions of error scenarios normative.

AxelNennker (changes requested):

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -279,6 +290,175 @@ Camara recommends that implementations run the OIDF interoperability suite and a
...
 * [RFC 8259 - The JavaScript Object Notation (JSON) Data Interchange Format](https://www.rfc-editor.org/info/rfc8259)
 * [RFC 8414 - OAuth 2.0 Authorization Server Metadata](https://www.rfc-editor.org/info/rfc8414)
 
+## Appendix A: Error Responses
+
+This section describes the error responses that the Authorization Server MUST return for all flows defined in this profile. Based on the supported functionality, the set of errors and the scenarios that can generate them are defined from the specifications already referenced in this document.
+
+### Authentication Error Response
+
+#### Authorization Code Flow
```
trehman-gsma, line 303: I think this new section is valuable, particularly for implementor technical teams we've spoken to that are not so close to OpenID specification, and for integrators who can use it as a concise requirements guide. I think it would be helpful if the sub-sections linked out to the relevant OIDC Error specification for readers to review the full context of the expected error response (e.g. [Authentication Error Response](https://openid.net/specs/openid-connect-core-1_0.html#AuthError)). For example, in Auth Code flow, there are other parameters that can be returned (e.g. a state parameter may be required depending on the input).

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -279,6 +290,175 @@ Camara recommends that implementations run the OIDF interoperability suite and a
...
+
+If the request fails due to a missing, invalid, or mismatching redirection URI, or if the client identifier is missing or invalid,
+the authorization server MUST NOT automatically redirect the user-agent and SHOULD inform the resource owner of the
```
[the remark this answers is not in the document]
garciasolero: As far as I know, the only suggestion ever made was to send the request object by value, not by reference. So this error would not occur as it would return `request_uri_not_supported`.

documentation/CAMARA-Security-Interoperability.md
garciasolero: I added new links to relevant error specifications.

AxelNennker (changes requested):
Where does the "SHOULD inform the resource owner of the error" come from?
That mean that the authorization server returns a HTTP 200 OK and the user agent shows the error page that "informs" the user, right? 
What are the pros and cons to returning a HTTP 400 BAD REQUEST?
```
HTTP/1.1 400 Bad Request
  Content-Type: application/json
  Cache-Control: no-store

  {
   "error": "invalid_request_uri"
  }
  ```

File: documentation/CAMARA-Security-Interoperability.md [hunk 3]
```diff
@@ -279,6 +290,173 @@ Camara recommends that implementations run the OIDF interoperability suite and a
...
 
+## Appendix A: Error Responses
+
+This section describes the error responses that the Authorization Server MUST return for all flows defined in this profile. Based on the supported functionality, the set of errors and the scenarios that can generate them are defined from the specifications already referenced in this document.
+
+### Authentication Error Response
+
+#### Authorization Code Flow
+
+If the request fails due to a missing, invalid, or mismatching redirection URI, or if the client identifier is missing or invalid, the authorization server MUST NOT automatically redirect the user-agent and SHOULD inform the resource owner of the error. For instance, the authorization server MAY display a message to the user describing the problem.
```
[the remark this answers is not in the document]
AxelNennker: I copied the text from the standard. I think that is safer, because otherwise we might inadvertently change the standard. Specifications must be precise and reformulating specifications bears the danger of introducing misunderstandings and interpretations. The previous sentence contained the adjective "automatically" which some implementer could understand as "if I configure the authorization server" then that is not "automatic" because I configured it manually. Maybe that is contrived, but so is interpreting the specified tel:" prefix for login_hint as "Tel:" being allowed. Whatever.
 
I suggest that we copy the error response definitions verbatim to this Appendix.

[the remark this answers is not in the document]
garciasolero: The text is exactly as it is in the [OAuth 2.0 specification](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1), with the order of the sentence changed. The only addition, as you suggested, was the new sentence regarding how the user should be informed.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -279,6 +290,173 @@ Camara recommends that implementations run the OIDF interoperability suite and a
...
+
+
+|         Error Code          | Scenario                                                                                                                                                                                                                                                                                                                                      |
+|:---------------------------:|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+|      `invalid_request`      | The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed.<br/>If the server requires PKCE and the client does not send the `code_challenge` in the request.<br/>If the server supporting PKCE does not support the requested `code_challenge_method`. |
```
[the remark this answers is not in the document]
garciasolero: I'm okay with adding the reference to the PKCE standard. However, I don't see why there should be any reference to `error_description` or `error_uri` in that section, as they could equally apply to other error codes. It is assumed that the authserver should use the `error_description` or the response of `error_uri` to explain the nature of the errors.

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
AxelNennker: I looked at OIDC https://openid.net/specs/openid-connect-core-1_0.html#AuthError and copied the text from there.

Regarding "inform", if the user agent is a browser and the AZ returns HTTP 200 OK "information" then the resource owner would be informed. Or the AZ redirects to location: https://errors.mno.com/invalid_request_uri.
If the AZ returns HTTP 400 Bad Request { "error": "invalid_request_uri" } is the user then informed?
If the API Consumer has a mobile app, then how would it react to these cases?

I think HTTP 400 Bad Request is the cleanest error response if the redirect_uri is invalid.
Also that is my understanding of the standards OIDC and OAuth2.

I would remove the "inform" sentence.

Further reading on redirect_uri validation etc: 
https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#:~:text=stage%2C%20maybe%20because-,it,-doesn%27t%20seem%20to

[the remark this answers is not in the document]
AxelNennker: I would always quote the higher standard, meaning OIDC is based on OAuth2. So, I would quote OIDC first.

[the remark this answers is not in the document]
garciasolero: Your strategy of referencing the higher standard may be good in many cases, but in this case the OIDC text is less precise and does not cover the case of the missing redirect_uri.

In the authserver cannot redirect scenario, the standard is clear that you must inform the resource owner, i.e. the user. So you need to communicate with them in a language they understand by displaying a message on a natural language page. Although it may seem strange to us, a human does not understand HTTP status codes, JSON format or OAuth error codes. ;)

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -279,6 +290,175 @@ Camara recommends that implementations run the OIDF interoperability suite and a
 * [RFC 7519 - JSON Web Token (JWT)](https://www.rfc-editor.org/info/rfc7519)
 * [RFC 7521 - Assertion Framework for OAuth 2.0 Client Authentication and Authorization Grants](https://www.rfc-editor.org/info/rfc7521)
 * [RFC 7523 - JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants](https://www.rfc-editor.org/info/rfc7523)
+* [RFC 7636 - Proof Key for Code Exchange by OAuth Public Clients](https://www.rfc-editor.org/info/rfc7636)
 * [RFC 8259 - The JavaScript Object Notation (JSON) Data Interchange Format](https://www.rfc-editor.org/info/rfc8259)
 * [RFC 8414 - OAuth 2.0 Authorization Server Metadata](https://www.rfc-editor.org/info/rfc8414)
 
+## Appendix A: Error Responses
```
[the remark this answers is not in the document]
AxelNennker: Everything in this specification is normative except examples.

I would like to have this sentence in the Appendix 
> For the convenience of implementers this section is a collection of error responses defined in CIBA, OIDC and OAuth2. This Appendix does not change the definition of error responses specified in CIBA, OIDC and OAuth2.

jpengar (dismissed):
LGTM

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
AxelNennker: @garciasolero 
I don't understand what an invalid request_uri has to do with whether a request object is send by value or reference.

Common invalid redirect_uri cases are:
- value is not matching an registered values for the client_id
  e.g. redirect_uri=https://attacker.example.com/ with a registered value being https://client.example.com/
  e.g. redirect_uri=ts43client: with a registered value being https://ts43client.example.com/
In these cases the OP MUST not use https://attacker.example.com/ or ts43client in any way but must respond with an HTTP 400 Bad Request

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
AxelNennker: I think we need to make sure that the new Appendix does not change the standards.

> For the convenience of implementers this section is a collection of error responses defined in CIBA, OIDC and OAuth2. This Appendix does not change the definition of error responses specified in CIBA, OIDC and OAuth2.

If we change or clarify the standards for .e.g. interoperability reasons then I would prefer to have that change in the main body of the specification. When we create a long list of something that is already standardized then we are creating a document that is hard to read and it becomes hard for the operator of the OpenId Provider to make sure that they follow the profile. If an operator buys an openid provider of the shelf then they can be quite certain that the OpenId Provider implements the standards. If it does not, than that must be corrected. CAMARA authorization servers MUST implement the OpenId Connect standard (and CIBA). The OpenId Foundation recommends to use [certified OpenId Providers](https://openid.net/certification/). Even major Internet players and Identity companies got their OpenId Provider implementation wrong which led to security issues. It is a major undertaking to implement an OpenId provider.

jpengar:
@AxelNennker and everyone, just checking in to see if you're okay with the current PR content. There hasn't been any activity recently, so could we agree to merge the PR as is?

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -279,6 +290,174 @@ Camara recommends that implementations run the OIDF interoperability suite and a
...
+|    `unauthorized_client`    | The client is not authorized to request an authorization code using this method.                                                                                                                                                                                                                                                                                                                                                                            |
+|       `access_denied`       | The user or authorization server denied the request (for example, the user cannot be authenticated or denied the consent).                                                                                                                                                                                                                                                                                                                                  |
```
mhfoo, line 314: If the IP address cannot be mapped to a phone number, is `access_denied` the correct error code to be returned?

The example scenario: 
If the source IP address is routed to the wrong home operator and the home operator is unable to find the corresponding mobile number, to perform network authentication. Should the error code be `access_denied` or another error code?

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -2,35 +2,46 @@
...
+  * [Missing "openid" scope](#missing-openid-scope)
+  * [Purpose](#purpose)
+    * [Purpose as a scope](#purpose-as-a-scope)
+    * [Outlook on purpose-handling leveraging Rich Authorization Request](#outlook-on-purpose-handling-leveraging-rich-authorization-request)
+  * [ID Token](#id-token)
+    * [ID Token sub claim](#id-token-sub-claim)
+  * [Client Authentication](#client-authentication)
+  * [OpenId Foundation Certification](#openid-foundation-certification)
+  * [References](#references)
+  * [Appendix A: Error Responses](#appendix-a-error-responses)
```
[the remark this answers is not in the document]
jpengar:
```suggestion
  * [Appendix A (Normative): Error Responses](#appendix-a-normative-error-responses)
```

jpengar (approved):
LGTM

documentation/CAMARA-Security-Interoperability.md
jpengar: Auth Code flow -> access_denied
CIBA -> unknown_user_id

jpengar:
> @AxelNennker and everyone, just checking in to see if you're okay with the current PR content. There hasn't been any activity recently, so could we agree to merge the PR as is?

A friendly reminder. I would suggest setting this week as the deadline.

sebdewet (approved):
LGTM

eric-murray (approved):

jpengar:
@AxelNennker Please review this PR to unblock and potentially merge it. The latest feedback from WG is available at https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/56754234/2024-12-18+ICM+Minutes

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
AxelNennker: OIDC is more precise "If the Redirection URI is invalid, the Authorization Server MUST NOT redirect the User Agent to the invalid Redirection URI. " While OAuth2 says MUST not automatically, which would allow an error page with a button which the user might click and be tricked into a malicious URL.

My "suggestion" above follows OIDC 3.1.26 and is more strict.

AxelNennker (changes requested):

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
garciasolero: Sorry, Axel, but I don't understand what you mean. Initially, you talked about the `invalid_request_uri` error related to passing the request object by reference (`request_uri` parameter), and now you're talking about the `redirect_uri` parameter, which is unrelated.

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
garciasolero: If the main concern is with the use of the word "automatically," it can be removed from the text. However, I think the rest of the paragraph in my proposal is more comprehensive as it includes the missing and mismatching `redirect_uri` parameter.

jpengar:
> OIDC is more precise "If the Redirection URI is invalid, the Authorization Server MUST NOT redirect the User Agent to the invalid Redirection URI. " While OAuth2 says MUST not automatically, which would allow an error page with a button which the user might click and be tricked into a malicious URL.
>
> My "suggestion" above follows OIDC 3.1.2.6 and is more strict.

Could just removing the word "automatically" do the trick? As suggested by @garciasolero in https://github.com/camaraproject/IdentityAndConsentManagement/pull/220#discussion_r1893641496
If this paragraph is the only issue, it should be easy to reach a final agreement on the wording to be used and merge the PR.

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
garciasolero: "automatically" removed.

jpengar (approved):
LGTM

jpengar (commented):
LGTM. @AxelNennker As discussed in yesterday's WG call, could you now check if we can merge the PR already?

jpengar (approved):

sebdewet (dismissed):

AxelNennker:
I fixed the merge conflicts by copying the table of contents from main adding the toc-entry for the Appendix A.

AxelNennker:
There seem to be two addtions in the PR that do not belong here:
1) the DPoP table entries seem be centered using whitespace. I have read somewhere that whitespace is ignored but it seems to work
image

2) PKCE entry in References

I am OK with these two changes, although the should be in separate PRs.

documentation/CAMARA-Security-Interoperability.md
AxelNennker: https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md does not say what error is returned if network-based authentication fails, I think. I have not found that error case. Network-based authentication that equated subscriber and end-user should define error responses.

If the user-agent is redirected to the end-user authentication page and the end-user clicks "cancel", then the authorization server should redirect to the redirect_url with error access_denied.
If the user-agent is redirected to the consent collection page and the end-user clicks something like "deny consent", then Authorization code flow returns access_denied. 

CIBA returns access_denied if `The end-user denied the authorization request. `.

AxelNennker (changes requested):

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -32,7 +32,8 @@
   * [Client Authentication](#client-authentication)
   * [OpenId Foundation Certification](#openid-foundation-certification)
   * [References](#references)
-   
+  * [Appendix A (Normative): Error Responses](#appendix-a-normative-error-responses)>
```
AxelNennker, line 35:
```suggestion
  * [Appendix A (Normative): Error Responses](#appendix-a-normative-error-responses)
```

AxelNennker (changes requested):

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -321,9 +322,177 @@ CAMARA recommends that implementations run the OIDF interoperability suite and a
...
+|    `unauthorized_client`    | The client is not authorized to request an authorization code using this method.                                                                                                                                                                                                                                                                                                                                                                            |
+|       `access_denied`       | The user or authorization server denied the request (for example, the user cannot be authenticated or denied the consent).                                                                                                                                                                                                                                                                                                                                  |
```
AxelNennker, line 349:
```suggestion
|       `access_denied`       | If the end-user was asked to authenticate, but end-user authentication was cancelled, then the `access_denied` MUST be returned.<br/>If the end-user was asked for their consent, but declined, then `access_denied` MUST be returned. <br/>If network-based authentication identified the subscriber, and the `prompt` parameter has the value of `none`, but the consent was revoked, then `access_denied` MUST be revokded.                                                                                                                                                                                                                                                                                                                                  |
```

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -321,9 +322,177 @@ CAMARA recommends that implementations run the OIDF interoperability suite and a
...
+|       `invalid_scope`       | The requested scope is either invalid, unknown, or malformed.                                                                                                                                                                                                                                                                                                                                                                                               |
+|     `consent_required`      | The authorization server requires user consent. This error MAY be returned when the `prompt` parameter value in the request is `none`, but the request cannot be completed without displaying a user interface for user consent.                                                                                                                                                                                                                            |
```
AxelNennker, line 352:
```suggestion
|     `consent_required`      | The authorization server requires user consent. This error MAY be returned when the `prompt` parameter value in the request is `none`, but the request cannot be completed without displaying a user interface for user consent.  This error is typically returned if the authorization server cannot obtain consent when the `prompt`parameter value is `prompt`.                                                                                                                                                                                                                    |
```

AxelNennker (changes requested):

documentation/CAMARA-Security-Interoperability.md
garciasolero: The third sentence is not correct and contains typos:

> none
    The Authorization Server MUST NOT display any authentication or consent user interface pages. An error is returned if an End-User is not already authenticated or the Client does not have pre-configured consent for the requested Claims or does not fulfill other conditions for processing the request. The error code will typically be login_required, interaction_required, or another code defined in [Section 3.1.2.6](https://openid.net/specs/openid-connect-core-1_0.html#AuthError). This can be used as a method to check for existing authentication and/or consent.
    
 Therefore, if prompt=none is sent and capturing consent is required , the error should be `consent_required`.

documentation/CAMARA-Security-Interoperability.md
garciasolero: `prompt` is not a valid value for `prompt` parameter.

documentation/CAMARA-Security-Interoperability.md
eric-murray: > Therefore, if prompt=none is sent and capturing consent is required , the error should be `consent_required`

Is the intention of defining this error response to get the client to re-try authorisation with `prompt=consent`?

There will be (at least) two error scenarios when `prompt=none` - those where the authorisation server can proceed with suitable interaction (i.e. consent gathering), and those where it cannot. Separate error messages should be defined for these two scenarios if both are to be supported by CAMARA so that the client knows what to do next.

AxelNennker: > The third sentence is not correct and contains typos:
I changed the suggestion. Are the typos now fixed?

documentation/CAMARA-Security-Interoperability.md
AxelNennker: I corrected that.
> consent
> The Authorization Server SHOULD prompt the End-User for consent before returning information to the Client. If it cannot obtain consent, it MUST return an error, typically consent_required.
https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest:~:text=typically%20consent_required

documentation/CAMARA-Security-Interoperability.md
AxelNennker: > There will be (at least) two error scenarios when `prompt=none` - those where the authorisation server can proceed with suitable interaction (i.e. consent gathering), and those where it cannot. Separate error messages should be defined for these two scenarios if both are to be supported by CAMARA so that the client knows what to do next.

Doesn't the proposed texts on access_denied and consent_required cover two cases?
> If network-based authentication identified the subscriber, and the `prompt` parameter has the value of `none`, but the consent was **revoked**, then `access_denied` MUST be returned.

The other error scenario for prompt=none:
> consent_required: The authorization server requires user consent. This error MAY be returned when the `prompt` parameter value in the request is `none`, but the request cannot be completed without displaying a user interface for user consent.

eric-murray: > Doesn't the proposed texts on access_denied and consent_required cover two cases?

The definition of `access_denied` does not really make it clear that that error message should be returned when `prompt=none`, end user consent is required but not in place, and the authorisation server has no in-band mechanism to collect that consent.

But if that is the defined error code, then client behaviour is clear:
`access_denied` means "go away"
`consent_required` means "retry with `prompt=consent`"

AxelNennker: consent_required means consent required. The API consumer could try the same request without "prompt=none" at some later point in time or they could use "prompt=consent".

OIDC authorization code flow is a redirect-based flow and when interaction is allowed then the authorization server should redirect to the operator's consent collection page. Regardless whether there is no prompt parameter or prompt=consent.
I assume that there must be a webpage where the [Operator's Consent Master](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#:~:text=If%20necessary%2C%20it%20will%20check%20in%20the%20Operator%27s%20Consent%20Master%20whether%20User%20Consent%20has%20already%20been%20given%20for%20this%20Application%2C%20Scope%20and%20Purpose%20(Steps%205%2D6)) is managed by the end-user? 
Or does the end-user have to call the operator's support hotline and opt-out or in?

> If necessary, it will check in the Operator's Consent Master whether User Consent has already been given for this Application, Scope and Purpose (Steps 5-6).
.
I assume that the authorization server redirects to that page if consent needs to be managed.

eric-murray: > I assume that there must be a webpage where the [Operator's Consent Master](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#:~:text=If%20necessary%2C%20it%20will%20check%20in%20the%20Operator%27s%20Consent%20Master%20whether%20User%20Consent%20has%20already%20been%20given%20for%20this%20Application%2C%20Scope%20and%20Purpose%20(Steps%205%2D6)) is managed by the end-user?

In the near-term, I think that might be an optimistic assumption for many API providers! But I take the point that in-band consent gathering is a possibility for some use cases.

garciasolero:
> There seem to be an addition in the PR that does not belong here: The DPoP table entries seem be centered using whitespace. I have read somewhere that whitespace is ignored but it seems to work 

When I made the commit, due to merge conflicts, the IDE auto-formatted the table because the columns were not aligned. With the correction, the table now appears correctly.

documentation/CAMARA-Security-Interoperability.md
garciasolero: I added the use case `prompt=consent` with a sentence structured the same as the scenario `prompt=none`.

eric-murray:
@jpengar
I've already approved

AxelNennker:
Not a crucial review, but...
This PR changes something that is does not need to change.

In the past we had TOC entries using different list-markers.
Level 1 toc entries were '*'
Level 2 toc entries were '+'
Level 3 toc entries were '-'

This PR is mixing markers now because it changed some `+` to `*`.
If fear that increases the likelihood of merge conflicts.

We should agree on one scheme which marker to use when, or e.g. always use e.g. `*`

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -334,9 +344,177 @@ CAMARA recommends that implementations run the OIDF interoperability suite and a
...
+* [RFC 7636 - Proof Key for Code Exchange by OAuth Public Clients](https://www.rfc-editor.org/info/rfc7636)
 * [RFC 8259 - The JavaScript Object Notation (JSON) Data Interchange Format](https://www.rfc-editor.org/info/rfc8259)
 * [RFC 8414 - OAuth 2.0 Authorization Server Metadata](https://www.rfc-editor.org/info/rfc8414)
 * [RFC 9449 - OAuth 2.0 Demonstrating Proof of Possession (DPoP)](https://www.rfc-editor.org/info/rfc9449)
 * [GSMA GSMA Authorization Server – Authenticator capabilities](https://www.gsma.com/newsroom/gsma_resources/asac-01-v1-0/)
 * [GSMA TS.43 Service Entitlement Configuration](https://www.gsma.com/newsroom/gsma_resources/ts-43-service-entitlement-configuration/)
 
+## Appendix A (Normative): Error Responses
+
+This section describes the error responses that the Authorization Server MUST return for all flows defined in this profile. Based on the supported functionality, the set of errors and the scenarios that can generate them are defined from the specifications already referenced in this document, such as CIBA, OIDC and OAuth 2.0. This appendix does not change the definition of error responses specified in these standards.
```
AxelNennker, line 356:
```suggestion
This section describes the error responses that the Authorization Server MUST return for all flows profiled by this document. Based on the supported functionality, the set of errors and the scenarios that can generate them are defined from the specifications already referenced in this document, such as CIBA, OIDC and OAuth 2.0. This appendix does not change the definition of error responses specified in these standards.
```

I argue that this profile does not "define" flows. The profile clarifies flows but it does not define flows.

AxelNennker (changes requested):

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -334,9 +344,177 @@ CAMARA recommends that implementations run the OIDF interoperability suite and a
...
+|    `unauthorized_client`    | The client is not authorized to request an authorization code using this method.                                                                                                                                                                                                                                                                                                                                                                            |
+|       `access_denied`       | The user or authorization server denied the request (for example, the user cannot be authenticated or denied the consent).                                                                                                                                                                                                                                                                                                                                  |
```
AxelNennker, line 371:
```suggestion
| `access_denied` | If the end-user was asked to authenticate, but end-user authentication was cancelled, then the `access_denied` MUST be returned.<br/>If the end-user was asked for their consent, but declined, then `access_denied` MUST be returned. <br/>If network-based authentication identified the subscriber, and the `prompt` parameter has the value of `consent`, but under the current legislation the end-user has no right to opt-out or otherwise revoke their consent, then `access_denied` MUST be returned. |
```

AxelNennker:
Sorry, this careful review takes so long. The appendix is normative and might therefore change the underlying standards e.g. by defining new flows etc.

Currently I have two remaining issues with this PR:
- make clear that the appendix does not define new flows
- mention all scenarios where access_denied are responded

@eric-murray said in another discussion that the sim-swap api makes no sense if the user can revoke their consent or opt-out. So, in cases where the user has no right to opt-out or revoke consent, access_denied is returned.

documentation/CAMARA-Security-Interoperability.md
garciasolero: @AxelNennker,

Please, can we merge the PR as it is now since there seemed to be consensus from most people? You are adding new error scenarios , of which it is not even clear that there is even a need to return the errors you propose. We could discuss this new scenarios in later issues/PRs.

If `prompt=consent` is sent, and there is no legal requirement to capture consent, the request could be processed without returning an error. If an error were returned, I doubt it would be an `access_denied` error, as the user is not involved.

jpengar (approved):

PR #221, opened 2024-10-23
Clarify case sensitivity of parameter names and values

eric-murray:
#### What type of PR is this?
* documentation

The CAMARA Security and Interoperability profile does not clarify whether parameter names and values are case sensitive.

It is proposed to adopt the [RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-1.9) convention that parameter names and values are case sensitive unless otherwise stated.

Fixes #

None

#### Changelog input

```
 release-note
 - Add document convention that parameter names and values are case-sensitive unless otherwise stated
```

#### Additional documentation 
None

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -51,10 +51,11 @@ By encapsulating these elements within this document, it aims to provide a compr
 
 The target audience for this document is the service/technical departments of operators exposing network functions via standard CAMARA APIs and the applications or client systems that consume CAMARA standard APIs to make use of the operator's network capabilities.
 
-
 ## Conventions
 
 The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.
+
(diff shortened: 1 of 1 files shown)


AxelNennker:
Specifications are specifications. If we define e.g. the "tel:" prefix for the login_hint format then that is the prefix.
No other interpretations. Neither "TEL:" nor "Tel:" nor "tEL:" are valid values because the spec specifies "tel:" and nothing else.

eric-murray:
> Specifications are specifications. If we define e.g. the "tel:" prefix for the login_hint format then that is the prefix. No other interpretations. Neither "TEL:" nor "Tel:" nor "tEL:" are valid values because the spec specifies "tel:" and nothing else.

Sure. And this statement, which other standards feel the need to include, **unambiguously clarifies this**.

garciasolero (approved):

AxelNennker:
I am trying to avoid repeating standards.
I have no idea how anybody can interpret the specification of e.g. login_hint "tel:" prefix that e.g "Tel:" is conforming to that specification.

I think the "clarification" this PR provides is very unneeded. You do not have to be an "expert" to read our spec and conclude that e.g. only "tel:" is a valid prefix.

If we go there that we start stating the obvious we end up with documents that are harder to understand because now e.g. OpenId Provider implementers have to read the OIDC core standard **plus** our document.

But, of course, if the group feels it helps then we can add this one sentence.

jpengar (approved):

AxelNennker (approved):

PR #226, opened 2024-11-12
recommend auth code flow using signed requests

### LINKED ISSUE #205: Spring25: Proposal to RECOMMEND the use of Signed Request Object for the /authorize endpoint to prevent abuse
**Problem description**
The simplest version of the OIDC `/authorize` endpoint is an unprotected endpoint with no means to authenticate the direct and indirect API Consumer. Anybody can monitor the existing traffic interaction with the `/authorize` endpoint (using a frontend application), and then perform a brute force attack on the /authorize endpoint to generate unnecessary auth codes.
The `PKCE` and `nonce` does not prevent the above scenarios.

In a federated scenario, the simplest version of the OIDC `/authorize` endpoint will allow chaining of unnecessary calls to downstream OIDC `/authorize` endpoints.

This is reopening issue #128 for Spring25 consideration.

**Possible evolution**
For the Spring25 CAMARA meta-release:

Recommend Signed Request Object for `/authorize` endpoint to be implemented, as per [OIDC, Chapter 6, Passing Request Parameters as JWTs](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The application backend will prepare and sign the request object using a private key, with the corresponding public key being published on well-know jwks endpoint. The application backend will then pass the signed request object to the application front using a 302 redirect response, where the application frontend will follow the redirect URL to invoke the intended `/authorize` endpoint. The `/authorize` endpoint will validate the signature of the signed request object with the corresponding jwks endpoint (which was registered as part of the onboarding process), identified using the client_id.

**Alternative solution**

**Additional context**
Related to [Number Verification PR 94](https://github.com/camaraproject/NumberVerification/pull/94)

AxelNennker:
#### What type of PR is this?

* documentation

This PR recommends using signed OIDC authoritzation code flow requests.

Fixes #205

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -9,6 +9,7 @@
     + [Transport Security](#transport-security)
     + [Sender-Constrained Tokens](#sender-constrained-tokens)
   * [OIDC Authorization Code Flow](#oidc-authorization-code-flow)
+    + [Signed Authentication Requests](#signed-authentication-requests)
     + [Optional Parameters](#optional-parameters)
     + [Cross-Site Request Forgery Protection](#cross-site-request-forgery-protection)
   * [Client-Initiated Backchannel Authentication Flow](#client-initiated-backchannel-authentication-flow)
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -205,7 +205,7 @@ The client MUST authenticate with the authorization server as described in [Clie
 ## The Scope Parameter
 
 
-Scope values determine the specific CAMARA services being requested by the Service Provider, ject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API projects. This document does not change OIDC definitions of scope values.
+Scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API projects. This document does not change OIDC definitions of scope values.
 
 
 ---
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -205,7 +205,7 @@ The client MUST authenticate with the authorization server as described in [Clie
 ## The Scope Parameter
 
 
-Scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API projects. This document does not change OIDC definitions of scope values.
+Scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API subprojects. This document does not change OIDC definitions of scope values.
 
 
 ---
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -66,7 +66,7 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
-It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication).
+It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication). This document recommends in the [client credentials section](#client-authentication) that the `aud` value SHOULD be the URL of the Authorization Server's [Token Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint). The same recommendation is given for the `aud` value of the signed request object.
 
 ### Cross-Site Request Forgery Protection
 
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -77,6 +77,8 @@ The authorization server MUST check the value of the `aud` field and reject sign
 
 Note: Care must be taken in a multi-tenant environment that a signed authentication request for one tenant is not accepted at another tenant endpoint.
 
+Note: For security reasons it is recommended that the API consumer never includes a `sub` field in the signed request object, because otherwise the signed request object might be used for client authenticaton.
+
 ### Cross-Site Request Forgery Protection
 
 CAMARA REQUIRES cross-site request forgery (CSRF) protection.
@@ -203,13 +205,13 @@ The client MUST authenticate with the authorization server as described in [Clie
 ## The Scope Parameter
 
 
-Scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API subprojects. This document does not change OIDC definitions of scope values.
+Scope values determine the specific CAMARA services being requested by the Service Provider, ject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API projects. This document does not change OIDC definitions of scope values.
 
 
 ---
 **NOTE**
 
-Scope values are an integral part of any OAuth2 and OIDC implementation. The API implementation enforces API access based on scope (if the CAMARA API subproject defines scopes).
+Scope values are an integral part of any OAuth2 and OIDC implementation. The API implementation enforces API access based on scope (if the CAMARA API project defines scopes).
 Therefore scopes should be available to API implementations.
 
 ---
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -66,6 +66,8 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
+It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for client authentication.
+
 ### Cross-Site Request Forgery Protection
 
 CAMARA REQUIRES cross-site request forgery (CSRF) protection.
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -8,6 +8,7 @@
    * [General Considerations](#general-considerations)
       + [Transport Security](#transport-security)
    * [OIDC Authorization Code Flow](#oidc-authorization-code-flow)
+      + [Signed Authentication Requests](#signed-authentication-requests)
       + [Cross-Site Request Forgery Protection](#cross-site-request-forgery-protection)
    * [Client-Initiated Backchannel Authentication Flow](#client-initiated-backchannel-authentication-flow)
       + [Optional Parameters](#optional-parameters)
@@ -66,7 +67,15 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
-It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication). This document recommends in the [client credentials section](#client-authentication) that the `aud` value SHOULD be the URL of the Authorization Server's [Token Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint). The same recommendation is given for the `aud` value of the signed request object.
+### Signed Authentication Requests
+
+It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication). 
+
+It is RECOMMENDED that the value of the `aud` field of the signed authentication request is the URL of the [Authorization Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint).
+The authorization server MAY accept different values of the `aud` field e.g. the `issuer` field of its [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata). 
+The authorization server MUST check the value of the `aud` field and reject signed authentication requests if the value of the `aud` is not associated with the authorization server.
+
+Note: Care must be taken in a multi-tenant environment that a signed authentication request for one tenant is not accepted at another tenant endpoint.
 
 ### Cross-Site Request Forgery Protection
 
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -211,7 +211,7 @@ Scope values determine the specific CAMARA services being requested by the Servi
 ---
 **NOTE**
 
-Scope values are an integral part of any OAuth2 and OIDC implementation. The API implementation enforces API access based on scope (if the CAMARA API project defines scopes).
+Scope values are an integral part of any OAuth2 and OIDC implementation. The API implementation enforces API access based on scope (if the CAMARA API subproject defines scopes).
 Therefore scopes should be available to API implementations.
 
 ---
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -66,7 +66,7 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
-It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for client authentication.
+It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication).
 
 ### Cross-Site Request Forgery Protection
 


documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,8 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
+It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for client authentication.
```
jpengar, line 69:
```suggestion
It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication).
```

mhfoo:
My apologies, I created #233, before seeing this PR. Possible to use #233 instead?

garciasolero:
I would like to clarify why the OIDC specification was chosen over RFC 9101. It is important because, although they are very similar, they have small but significant differences.

[OIDC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject):

> When the request parameter is used, the OpenID Connect request parameter values contained in the JWT supersede those passed using the OAuth 2.0 request syntax. However, parameters MAY also be passed using the OAuth 2.0 request syntax even when a Request Object is used; this would typically be done to enable a cached, pre-signed (and possibly pre-encrypted) Request Object value to be used containing the fixed request parameters, while parameters that can vary with each request, such as state and nonce, are passed as OAuth 2.0 parameters. 

[RFC 9101](https://www.rfc-editor.org/rfc/rfc9101.html#section-5):

> The client MAY send the parameters included in the Request Object duplicated in the query parameters as well for backward compatibility, etc.  However, the authorization server supporting this specification MUST only use the parameters included in the Request Object.

From my point of view, it is more interesting to follow RFC 9101, as it would ignore parameters that the client did not have to send and could potentially change how an authserver processes the request.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,8 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
+It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication). This document recommends in the [client credentials section](#client-authentication) that the `aud` value SHOULD be the URL of the Authorization Server's [Token Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint). The same recommendation is given for the `aud` value of the signed request object.
```
garciasolero, line 69: Both in [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject) and [RFC 9101](https://www.rfc-editor.org/rfc/rfc9101.html#section-4), the valid value for the `aud` field should be the issuer of the authorization server. I don't see any specification that recommends the Token Endpoint for signed request object.

Even in the [CIBA](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1.1) specification, it is also established that the valid value should be the issuer of the authorization server.

eric-murray: @garciasolero
The CAMARA specification [specifically](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-authentication) makes this recommendation.

However, this applies to client authentication, and for authorisation code flow, it is only the `/token` endpoint where the client is authenticated. For signing the authorisation request, I agree that is does not necessarily follow that the `aud` should also be the URL of the token endpoint.

garciasolero: Yes, you are right, hence my comment. The recommendations apply depending on what type of assertion it is, and we should be aligned with what the standards say.

mhfoo: [OIDC Core](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests) refers to [\[RFC9101\]](https://www.rfc-editor.org/rfc/rfc9101.html#name-explicit-endpoints), under section [Explicit Endpoints](https://www.rfc-editor.org/rfc/rfc9101.html#name-explicit-endpoints), it is RECOMMENDED good practice to explicitly state the intended interaction endpoints, in this context the Authorization endpoint (authorization_endpoint)

mhfoo:
```suggestion
It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication). 

This document recommends the value of `aud` (Audience) claim MUST be the Authorization endpoint as RECOMMENDED by [\[RFC9101\]](https://www.rfc-editor.org/rfc/rfc9101.html#name-explicit-endpoints).
```

mhfoo:
I would like to propose to append the following point regarding the `sub` claim for signed Request Object, as there are cases of implementation, the `sub` is wrongly included.

8< ------------------------
The `sub` (Subject) claim in the signed Request Object MUST NOT be present, to prevent [cross-JWT confusion](https://www.rfc-editor.org/rfc/rfc9101.html#name-cross-jwt-confusion).

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,8 @@ All network connections MUST use TLS 1.2 or better.
...
+
```
mhfoo, line 70:
```suggestion
The sub (Subject) claim in the signed Request Object MUST NOT be present, to prevent [cross-JWT confusion](https://www.rfc-editor.org/rfc/rfc9101.html#name-cross-jwt-confusion).

```

mhfoo (changes requested):
Added proposed changes, including references

mhfoo:
> * `aud` value. Telefónica believes that the `aud` field should be the issuer of the authorization server as per standards (both [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject) and [RFC 9101](https://www.rfc-editor.org/rfc/rfc9101.html#section-4))

Referring to the `aud` value statement in [ODIC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject):
"The aud value SHOULD be **_or_** include the OP's Issuer Identifier URL."

Breaking it down:
a) The aud value SHOULD be the OP's Issuer Identifier URL, or
b) The aud value SHOULD include the OP's Issuer Identifier URL.

Point b) is read as "OP's Issuer Identifier URL" is used as a prefix.

mhfoo:
> * `sub` claim requirement to avoid cross-JWT confusion.

Referring to [OIDC, Client Authentication, private_key_jwt](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication), the following statement is present:
"The JWT MAY contain other Claims. Any Claims used that are not understood MUST be ignored."

If this signed request object (for /authorize) `aud` value is the OP's Issuer Identifier URL and the 'sub' happens to be populated with the client_id of the OAuth Client, this could be used as the `client_assertion` for the /token or bc-authorize endpoint.

jpengar:
> > * `aud` value. Telefónica believes that the `aud` field should be the issuer of the authorization server as per standards (both [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject) and [RFC 9101](https://www.rfc-editor.org/rfc/rfc9101.html#section-4))
> 
> Referring to the `aud` value statement in [ODIC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject): "The aud value SHOULD be **_or_** include the OP's Issuer Identifier URL."
> 
> Breaking it down: a) The aud value SHOULD be the OP's Issuer Identifier URL, or b) The aud value SHOULD include the OP's Issuer Identifier URL.
> 
> Point b) is read as "OP's Issuer Identifier URL" is used as a prefix.

[RFC9101](https://www.rfc-editor.org/rfc/rfc9101.html#section-4) states: "The value of aud should be the value of the authorization server (AS) issuer..." in that sense. So again it sounds reasonable to suggest that the aud field should be the issuer of the authorisation server.

documentation/CAMARA-Security-Interoperability.md
AxelNennker: @mhfoo https://openid.net/specs/openid-connect-core-1_0.html#RequestObject does not mention `sub`.

> It represents the request as a JWT whose Claims are the request parameters specified in [Section 3.1.2](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint). This JWT is called a Request Object.

I would like to "resolve" this conversation by closing it.

documentation/CAMARA-Security-Interoperability.md
AxelNennker: I changed the text and hopefully improved it, so that this conversation be closed as resolved.

AxelNennker:
I changed the text regarding "aud".
I think the conversation regarding "sub" can be closed because sub is not part of the request object.

>  It represents the request as a JWT whose Claims are the request parameters specified in [Section 3.1.2](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint). This JWT is called a Request Object.

mhfoo:
> I think the conversation regarding "sub" can be closed because sub is not part of the request object.

@AxelNennker 
I would like to suggest adding a "Note" to mention `sub` claim should not be part of the request object.

documentation/CAMARA-Security-Interoperability.md
AxelNennker: I added a security note.

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
AxelNennker: Why should we use a separate standard when OIDC already defines [signed request objects](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject)?

AxelNennker:
For easier review: The current text is this. Can this be merged? I think it resolves all review comments and suggestions.

Signed Authentication Requests

It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](https://github.com/camaraproject/IdentityAndConsentManagement/blob/70b99742c5baffa8f263c1b04b8a459cb19feb1b/documentation/CAMARA-Security-Interoperability.md#client-authentication).

It is RECOMMENDED that the value of the aud field of the signed authentication request is the URL of the [Authorization Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint). The authorization server MAY accept different values of the aud field e.g. the issuer field of its [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata). The authorization server MUST check the value of the aud field and reject signed authentication requests if the value of the aud is not associated with the authorization server.

Note: Care must be taken in a multi-tenant environment that a signed authentication request for one tenant is not accepted at another tenant endpoint.

Note: For security reasons it is recommended that the API consumer never includes a sub field in the signed request object, because otherwise the signed request object might be used for client authenticaton.

garciasolero:
> For easier review: The current text is this. Can this be merged? I think it resolves all review comments and suggestions.

@AxelNennker,

I have not received feedback from your side on the [comment ](https://github.com/camaraproject/IdentityAndConsentManagement/pull/226#issuecomment-2490759026) regarding why the OIDC specification is preferred over RFC9101 when handling parameters outside the request object parameter.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -192,13 +205,13 @@ The client MUST authenticate with the authorization server as described in [Clie
 ## The Scope Parameter
 
 
-Scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API subprojects. This document does not change OIDC definitions of scope values.
+Scope values determine the specific CAMARA services being requested by the Service Provider, ject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API projects. This document does not change OIDC definitions of scope values.
```
shilpa-padgaonkar, line 208:
```suggestion
Scope values determine the specific CAMARA services being requested by the Service Provider, subject to the SP being registered to use those services. The scope values must be documented in the API OAS files by all CAMARA API projects. This document does not change OIDC definitions of scope values.
```

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -192,13 +205,13 @@ The client MUST authenticate with the authorization server as described in [Clie
...
 
 
 ---
 **NOTE**
 
-Scope values are an integral part of any OAuth2 and OIDC implementation. The API implementation enforces API access based on scope (if the CAMARA API subproject defines scopes).
+Scope values are an integral part of any OAuth2 and OIDC implementation. The API implementation enforces API access based on scope (if the CAMARA API project defines scopes).
```
shilpa-padgaonkar, line 214:
```suggestion
Scope values are an integral part of any OAuth2 and OIDC implementation. The API implementation enforces API access based on scope (if the CAMARA API subproject defines scopes).
```

documentation/CAMARA-Security-Interoperability.md
mhfoo: LGTM

mhfoo (commented):
LGTM, based on @shilpa-padgaonkar corrections.
@AxelNennker, Thanks.

mhfoo:
> I still don't understand why we change the `aud` value from what the standards say, both [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject) and [RFC 9101](https://www.rfc-editor.org/rfc/rfc9101.html#section-4). Even in the [CIBA](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1.1) specification states that the valid value should be the issuer of the authorization server, as also mentioned by @garciasolero [here](https://github.com/camaraproject/IdentityAndConsentManagement/pull/226#discussion_r1851839311).
> 
> And in the case of CIBA, [the profile doesn't even mention](https://github.com/camaraproject/IdentityAndConsentManagement/blob/c6b0e5bac227b1a9acabeb9e41dfd090be727a7c/documentation/CAMARA-Security-Interoperability.md?plain=1#L120) the `aud` claim at all.

@jpengar
Regarding the `aud` value, [RFC 9101, section 10.3, Explicit Endpoints](https://www.rfc-editor.org/rfc/rfc9101.html#section-10.3) recommends the following
>research such as [[BASIN](https://www.rfc-editor.org/rfc/rfc9101.html#BASIN)] points out that it is a good practice to explicitly state the intended interaction endpoints and the message position in the sequence in a tamper-evident manner so that the intent of the initiator is unambiguous. It is RECOMMENDED by this specification to use this practice for the following endpoints defined in [[RFC6749](https://www.rfc-editor.org/rfc/rfc9101.html#RFC6749)], [[RFC6750](https://www.rfc-editor.org/rfc/rfc9101.html#RFC6750)], and [[RFC8414](https://www.rfc-editor.org/rfc/rfc9101.html#RFC8414)]:
>(b) Authorization endpoint (authorization_endpoint)

In [RFC 8725 JSON Web Token Best Current Practices, section 3.12, Use Mutually Exclusive Validation Rules for Different Kinds of JWTs](https://datatracker.ietf.org/doc/html/rfc8725#preventing-confusion), the following is indicated:
>Each application of JWTs defines a profile specifying the required and optional JWT claims and the validation rules associated with them. If more than one kind of JWT can be issued by the same issuer, the validation rules for those JWTs MUST be written such that they are mutually exclusive, rejecting JWTs of the wrong kind. To prevent substitution of JWTs from one context into another, application developers may employ a number of strategies:
>- Use different "aud" values for different uses of JWTs from the same issuer. Then audience validation will reject JWTs substituted into inappropriate contexts.

The main security concern to address, is to prevent having a signed request object being used as a `client_assertion` in `private_key_jwt` [OIDC client authentication](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication), where it indicates the following for a `private_key_jwt`:
>The JWT MAY contain other Claims. Any Claims used that are not understood MUST be ignored.

By specifying the /authorize endpoint as the `aud` will mitigate this security concern.

documentation/CAMARA-Security-Interoperability.md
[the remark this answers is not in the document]
AxelNennker: OIDC says:
> Note that the Request Objects defined here are compatible with those specified by [The OAuth 2.0 Authorization Framework: JWT-Secured Authorization Request (JAR)](https://openid.net/specs/openid-connect-core-1_0.html#RFC9101) [RFC9101].
That is the only reference to RFC9101.

AxelNennker:
I think now all comments and suggestions are addressed. Please review again.
@jpengar @shilpa-padgaonkar @mhfoo @garciasolero @sebdewet

mhfoo (approved):
LGTM

eric-murray:
I'd like to see the conflicts resolved before approving

sebdewet (approved):
LGTM

jpengar:
I believe this PR (#226) was merged prematurely, as key concerns raised during the discussion remain unresolved. Specifically:

- The choice between OIDC and RFC9101 and the rationale for deviating from the latter.
- The handling of the aud parameter, which conflicts with existing specifications like CIBA.
- Questions about how to treat parameters not included in the signed request object, as RFC9101 specifies they must be ignored.

Since these points were still under discussion and the proposed text was not agreed upon by all participants, I recommend reverting this PR until a consensus is reached and the final text is approved.

@AxelNennker @mhfoo @garciasolero @sebdewet @eric-murray @shilpa-padgaonkar

AxelNennker:
I reverted the merge which created https://github.com/camaraproject/IdentityAndConsentManagement/pull/249

AxelNennker:
@garciasolero wrote:
> I would like to clarify why the OIDC specification was chosen over RFC 9101. It is important because, although they are very similar, they have small but significant differences.
> 
> [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#RequestObject):
> 
> > When the request parameter is used, the OpenID Connect request parameter values contained in the JWT supersede those passed using the OAuth 2.0 request syntax. However, parameters MAY also be passed using the OAuth 2.0 request syntax even when a Request Object is used; this would typically be done to enable a cached, pre-signed (and possibly pre-encrypted) Request Object value to be used containing the fixed request parameters, while parameters that can vary with each request, such as state and nonce, are passed as OAuth 2.0 parameters.
> 
> [RFC 9101](https://www.rfc-editor.org/rfc/rfc9101.html#section-5):
> 
> > The client MAY send the parameters included in the Request Object duplicated in the query parameters as well for backward compatibility, etc.  However, the authorization server supporting this specification MUST only use the parameters included in the Request Object.
> 
> From my point of view, it is more interesting to follow RFC 9101, as it would ignore parameters that the client did not have to send and could potentially change how an authorization server processes the request.

Sorry, @garciasolero, I thought this was addressed.

I think that 9101 and OIDC say the same thing regarding parameters existing both in the parameters and in the request object. 

For comparability reasons OAuth2 parameters that are required are also required in OIDC. But if parameters are also in the request object then those are the only ones that are relevant. 
OIDC expresses this using these words:  **When the request parameter is used, the OpenID Connect request parameter values contained in the JWT supersede those passed using the OAuth 2.0 request syntax.**
RFC9101 expresses this using these words: **the authorization server supporting this specification MUST only use the parameters included in the Request Object.**

You write: "From my point of view, it is more interesting to follow RFC 9101, as it would ignore parameters that the client did not have to send and could potentially change how an authorization server processes the request."

Maybe an some examples: (with line wraps within values for display purposes only)

#### Valid example. All required parameters are in the request AND the request object.
```
GET /authorize?
    request=base64urlencode(signed({
  "response_type": "code",
  "scope": "openid%20profile%20email",
  "client_id": "s6BhdRkqt3",
  "redirect_uri": "https%3A%2F%2Fclient.example.org%2Fcb"
})
    &response_type=code
    &scope=openid%20profile%20email
    &client_id=s6BhdRkqt3
    &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb HTTP/1.1
Host: server.example.com
```

#### Invalid OIDC example because required **parameters** are missing although they provided in the request object
```
GET /authorize?
    request=base64urlencode(signed({
  "response_type": "code",
  "scope": "openid%20profile%20email",
  "client_id": "s6BhdRkqt3",
  "redirect_uri": "https%3A%2F%2Fclient.example.org%2Fcb"
}) HTTP/1.1
Host: server.example.com
```

This PR does not change the OIDC standard, it just RECOMMENDS signed request objects.

AxelNennker:
Regarding aud:

[CIBA](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1) writes:
> To address that ambiguity the Issuer Identifier of the OP SHOULD be used as the value of the audience. To facilitate interoperability, the OP MUST accept its Issuer Identifier, Token Endpoint URL, or Backchannel Authentication Endpoint URL as values that identify it as an intended audience. 

We [recommend](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-authentication):
This document RECOMMENDS that for OIDC CIBA the audience SHOULD be the [Backchannel Authentication Endpoint](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_backchannel_endpoint).

This PR writes:
> It is RECOMMENDED that the value of the `aud` field of the signed authentication request is the URL of the [Authorization Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint).

This PR does what RFC9101 recommends, that is, the aud value is the URL of the endpoint the request is send to. While allowing other values to be used as long as the Authorization Server validates that the used value for "aud" is itself.

For interoperability reasons CIBA and OIDC are quite lenient regarding aud values. The responsibility in both flows lies with the authorization server. The profile recommends what rfc9101 recommends.

I thought that the above was the agreement here. Sorry, if I misinterpreted that.

jpengar:
> Regarding aud:
> 
> [CIBA](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1) writes:
> 
> > To address that ambiguity the Issuer Identifier of the OP SHOULD be used as the value of the audience. To facilitate interoperability, the OP MUST accept its Issuer Identifier, Token Endpoint URL, or Backchannel Authentication Endpoint URL as values that identify it as an intended audience.
> 
> We [recommend](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-authentication): This document RECOMMENDS that for OIDC CIBA the audience SHOULD be the [Backchannel Authentication Endpoint](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#auth_backchannel_endpoint).
> 
> This PR writes:
> 
> > It is RECOMMENDED that the value of the `aud` field of the signed authentication request is the URL of the [Authorization Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint).
> 
> This PR does what RFC9101 recommends, that is, the aud value is the URL of the endpoint the request is send to. While allowing other values to be used as long as the Authorization Server validates that the used value for "aud" is itself.
> 
> For interoperability reasons CIBA and OIDC are quite lenient regarding aud values. The responsibility in both flows lies with the authorization server. The profile recommends what rfc9101 recommends.
> 
> I thought that the above was the agreement here. Sorry, if I misinterpreted that.

@AxelNennker As was mentioned by @garciasolero [here](https://github.com/camaraproject/IdentityAndConsentManagement/pull/226#discussion_r1852169393), the recommendations apply depending on the type of assertion, client authentication assertion (client_assertion) or signed request object assertion.

The CAMARA profile specifies ["This document RECOMMENDS that for OIDC Authorisation Code Flow and OAuth2 Client Credentials Grant the audience SHOULD be the URL of the Authorisation Server's Token Endpoint. This document RECOMMENDS that for OIDC CIBA the audience SHOULD be the Backchannel Authentication Endpoint"](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-authentication), but that is for client authentication. Also, for the Auth code flow, there is no client authentication in the /authorize endpoint (only in /token). 
And the CAMARA profile document, which refers to signed request objects, in the case of CIBA (it does not even mention the aud parameter). So it relies on what the standard says, which is [The Audience claim MUST contain the value of the Issuer Identifier for the OP, which identifies the Authorization Server as an intended audience."](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1.1). So, we would be recommending different things to Auth code flow and CIBA.

Issue #215, opened 2024-10-12
Authentication method in authorization code flow


shilpa-padgaonkar:
**Problem description**
While discussing another [issue](#191 ), it was found that ICM participants have different interpretations on whether the CAMARA-API-access-and-user-consent.md and CAMARA-Security-Interoperability.md documents mandate an authentication method in authorization code flow to network based authentication or not.

**Expected action**
Clarify and get a consensus in the working group.

**Additional context**

AxelNennker:
DT's position on authentication by the API provider in Camara is that the API provider is in control of user authentication. 

ICM created a (security and) interoperability profile that exactly specifies what API Consumers and API producers need to agree upon to create an interoperable solution. If something is not specified in the interoperability profile, then we are back to OIDC, CIBA and OAuth2. 

All Camara subprojects leave authentication to the authorization server. Notable exception is the NumberVerify API which demands network-based authentication.

Under GDPR consent must be freely given, specific, informed, and unambiguous. 
It is DT's position that consent cannot be given "silently" with user authentication being only network-authentication. The API provider must authenticate the user, otherwise we do not know who the user is. The API provider must "inform" the user so that the consent can be freely given, specific, informed and unambiguous. There must be a user interaction otherwise the consent cannot be informed. 

DT's position on consent collection is that we do not allow delegated consent. The consent must be collected by the API provider.

shilpa-padgaonkar:
@jesus Thanks for your comments

I don’t believe the claim is that network-based authentication in the auth code flow is not documented in the CAMARA-API-access-and-user-consent.md. Rather, I want to emphasize that this does not automatically rule out non-network-based authentication.

For example, we have clearly stated that only poll mode is permitted, even though CIBA allows clients to use three modes. Similarly, we state that clients should not use the acr_values parameter. IMHO, if CAMARA intends to restrict something permitted by the standard, it should be explicitly stated.

However, the more important question from my side is, if an Application Service Provider (ASP) is willing to use the auth code flow with authentication methods other than network-based, and if the Mobile Network Operator (MNO) can support this, what are the reasons for CAMARA to forbid it?

eric-murray:
Hi @shilpa-padgaonkar

So my understanding is that this is only about authorisation code flow? Or does this issue include CIBA?

To me, the questions to be addressed are:
- Does the "CAMARA Security and Interoperability Profile" document currently allow other authentication mechanisms?
[...]

PR #228, opened 2024-11-14
Controlling User Interaction

AxelNennker:
#### What type of PR is this?

Add one of the following kinds:
* documentation

#### What this PR does / why we need it:
This PR repeats the CIBA and OIDC standards thus clarifying e.g. that prompt=none should be used if the API Consumer does not want user interaction.
User Interaction is controlled in OIDC by the authentication request parameter `prompt` which is mandatory to implement.

#### Which issue(s) this PR fixes:
Fixes #215

PR #229, opened 2024-11-14
Consistent User Authentication and Consent Collection

AxelNennker:
#### What type of PR is this?

* documentation

It is up to the operator to make a decision when consent collection is needed or not, based on the scope(s)/purpose declared by the API client and aligned with local legislation, ensuring that all operators under the same regulatory framework adopt a **consistent** approach.

This PR (hopefully) defines that consistent approach.

Update removed  #215 

Update: Please see https://github.com/camaraproject/IdentityAndConsentManagement/pull/228

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -70,9 +70,14 @@ The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.n
 
 It is up to the operator to make a decision when consent collection is needed or not, based on the scope(s)/purpose declared by the API client and aligned with local legislation, ensuring that all operators under the same regulatory framework adopt a *consistent* approach.
 
-This document defines that if the optional prompt parameter is absent from the OIDC authentication request then the API consumer uses network-based authentication to identify the subscription. 
-* If the subscription belongs to a business contract then the authorization server MUST return an error.
-* If the subscription belongs to a private contract but there are several contracts to the contract holder then the authorization server MUST return an error.
+This document defines that if the optional prompt parameter is absent from the OIDC authentication request then the API consumer uses network-based authentication to identify the subscription. If the subscription identified by network-based authentication cannot assure that the end-user equals the subscriber then an error MUST be returned.
+
+Examples for cases where the authorization server cannot assue that the end-user equals the subscriber:
+```
+* A business owner has one contract that is associated with multiple MSISDNs
+  then the authorization server cannot assume that the end-user is the subscriber.
+* A family has several contracts which all have the same subscriber then the authorization server cannot assume that the end-user is the subscriber.
+```
 
 The API provider could start User-Interaction to resolve the above cases, but that likely leads to different behaviour by different API producers.
 


documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,26 @@ All network connections MUST use TLS 1.2 or better.
...
+
+This document defines that if the optional prompt parameter is absent from the OIDC authentication request then the API consumer uses network-based authentication to identify the subscription. 
+* If the subscription belongs to a business contract then the authorization server MUST return an error.
+* If the subscription belongs to a private contract but there are several contracts to the contract holder then the authorization server MUST return an error.
+* If the subscription identifies one MSISDN but other subscriptions for the same MSISDN exist then the authoriztoin server MUST return an error.
+
+The API provider could start User-Interaction to resolve the above cases, but that likely leads to different behaviour by different API producers.
+
+The document defines that if the optional prompt parameter is absent from the OIDC authentication request and network-based authentication identifies one subscription that belongs to one person and that person does not own other subscriptions with that API provider, then the API provider can assume that the subscriber equals the end-user.
```
jpengar, line 80: I disagree, see my comment above.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,26 @@ All network connections MUST use TLS 1.2 or better.
...
+The API provider could start User-Interaction to resolve the above cases, but that likely leads to different behaviour by different API producers.
+
+The document defines that if the optional prompt parameter is absent from the OIDC authentication request and network-based authentication identifies one subscription that belongs to one person and that person does not own other subscriptions with that API provider, then the API provider can assume that the subscriber equals the end-user. 
+The decicion whether consent needs to be collected or not is then determined by the API provider 
+* by determining the legal basis of the purpose e.g.: contract, legitimate_interest, consent, etc,
+* and then checking if consent is required for the legal basis of the purpose
+* and then checking whether consent was already granted or revoked out-of-band.
+
+If consent needs to be optained then the API provider optains it according to [OIDC section 3.1.2.4](https://openid.net/specs/openid-connect-core-1_0.html#Consent).
+
```
jpengar, line 87: I believe that both #228 and #229 are providing a significant amount of supplementary information to the profile document. As I previously stated, this may be an acceptable approach for example to clarify prompt parameter-related behaviour. **However, if we have already reached a specific conclusion in #215, would it not be more straightforward to resolve the issue by including that concise and clear conclusion in the documentation within the Auth Code flow section?**

jpengar (changes requested):

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,26 @@ All network connections MUST use TLS 1.2 or better.
...
+
```
jpengar, line 88: I believe this PR is adding a significant amount of supplementary information to the profile document, which, as previously stated, may be appropriate for instances such as clarifying prompt parameter behavior. However, if we reach a specific conclusion in #215, wouldn't it be more efficient to address the issue by simply including that concise and straightforward conclusion in the profile document?

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,26 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
+### Consistent User Authentication and Consent collection
+
+It is up to the operator to make a decision when consent collection is needed or not, based on the scope(s)/purpose declared by the API client and aligned with local legislation, ensuring that all operators under the same regulatory framework adopt a *consistent* approach.
+
+This document defines that if the optional prompt parameter is absent from the OIDC authentication request then the API consumer uses network-based authentication to identify the subscription.
```
jpengar, line 73: The use of network-based authentication is not dependent on the prompt parameter. I would simply reiterate the conclusion proposed in #215:
>- Currently, CAMARA only considers network-based authentication in the Authorization Code Flow. Therefore, access tokens are issued for the network authenticated identifier.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,26 @@ All network connections MUST use TLS 1.2 or better.
...
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
+### Consistent User Authentication and Consent collection
+
+It is up to the operator to make a decision when consent collection is needed or not, based on the scope(s)/purpose declared by the API client and aligned with local legislation, ensuring that all operators under the same regulatory framework adopt a *consistent* approach.
+
+This document defines that if the optional prompt parameter is absent from the OIDC authentication request then the API consumer uses network-based authentication to identify the subscription. 
+* If the subscription belongs to a business contract then the authorization server MUST return an error.
+* If the subscription belongs to a private contract but there are several contracts to the contract holder then the authorization server MUST return an error.
+* If the subscription identifies one MSISDN but other subscriptions for the same MSISDN exist then the authoriztoin server MUST return an error.
```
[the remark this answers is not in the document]
AxelNennker: The last point is about Multi-SIM. This list tries to capture the current understanding.
I moved that point/sentence to the CIBA section because the consistent API Provider behavior is to return an error if the device can't be determined by `login_hint=tel:` in the CIBA case.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,26 @@ All network connections MUST use TLS 1.2 or better.
 
 The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
 
+### Consistent User Authentication and Consent collection
+
+It is up to the operator to make a decision when consent collection is needed or not, based on the scope(s)/purpose declared by the API client and aligned with local legislation, ensuring that all operators under the same regulatory framework adopt a *consistent* approach.
+
+This document defines that if the optional prompt parameter is absent from the OIDC authentication request then the API consumer uses network-based authentication to identify the subscription. 
+* If the subscription belongs to a business contract then the authorization server MUST return an error.
```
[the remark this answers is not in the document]
AxelNennker: I'll change that. The issue is not so much whether the contract is a business contract but rather whether the end-user equals the subscriber. This PR tries to capture the current understanding and replace assumptions by specifications what the defined behavior is.
I asked TDG what they implemented and they said "business contract -> no access token" because usually the business contract has multiple MSISDNs and we do not want the employee with one of the MSISDN to act on behalf of the subscriber.

documentation/CAMARA-Security-Interoperability.md
```diff
@@ -66,6 +66,26 @@ All network connections MUST use TLS 1.2 or better.
...
+* If the subscription belongs to a private contract but there are several contracts to the contract holder then the authorization server MUST return an error.
```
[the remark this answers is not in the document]
AxelNennker: I rephrased that part. I think if the authorization server cannot determine that the end-user reasonably equals the subscriber then the authorization server cannot check whether consent was given or revoked by the end-user.
Consent can only be granted by a data-subject (end-user), and if we can't identify the data-subject, then we do not know anything about their consent. 

Example: A spouse has two contracts with the API provider. After receiving the authentication request the authorization server does network-based authentication and identifies the subscriber. The authentication server knows nothing whether the other spouse ,who is not the subscriber, consented to e.g. device location. Starting consent collection does not meaningfully work in this case because in general only the subscriber can be authenticated as the end-user the API provider usually created one account for the subscription but not two for each person. Assuming the API provider created two separate accounts, one for each person, then consent collection would work but in CAMARA we need consistent API provider behavior, so returning an error in this case is the (privacy-)safe option.

documentation/CAMARA-Security-Interoperability.md
jpengar: I strongly suggest that PR content be kept as concise as possible, focusing on clarifying the conclusion proposed in #215 and using the same wording. This is in line with the intention to reach an agreement on the conclusion in #215 comments. The proposed PR content introduces additional topics that will require more detailed discussion in ICM. These topics could be initiated in parallel in new issues, although it would be necessary to ascertain whether they can be included in Spring25. Otherwise, it will be challenging to merge this PR.

**If the WG considers that we have not yet reached a final conclusion in #215, this should be addressed as a priority.**

AxelNennker:
I removed Fixes #215 and rephrased. The PRs #228 and #229 were started be cause of #215. I wanted to write the consensus down while not excluding operator token and clarify prompt behavior.

I want to achieve that we have a consistent behavior by all API producers.
I think that end-user not-equal subscriber leads to GDPR problems, and therefore wrote that we should fail in the privacy-safe side.

documentation/CAMARA-Security-Interoperability.md
AxelNennker: This PR is an attempt to come to a conclusion in #215 by specifying behavior of the authorization server that is interoperable.

OIDC provides everything we need.
- the API consumer can test the water with `prompt=none` and if the response is "consent_needed" the API consumer can start OIDC consent collection with `prompt=consent`.

documentation/CAMARA-Security-Interoperability.md
AxelNennker: There a some APIs that explicitly require the device to be identified like e.g. NumberVerification.
There a some  APIs that do not restrict their use cases in this way e.g. KYC-fill-in

We want to make CAMARA APIs useful and attractive to API consumers. If we specify how authorization servers react to requests, then we get an interoperable solution that covers more use cases.

jpengar:
@AxelNennker Should we consider closing these old PRs, which correspond to previous discussions that have already been closed or resolved? It would be good to tidy up the list of issues and PRs.

Issue #232, opened 2024-11-19
Allow to use operator token for device authentication in OpenID Auth code flow


Elisabeth-Ericsson:
**Problem description**
With issue #145 it was agreed to allow using the operator token for device authentication by passing this token as value in the /bc_authorize request as 3rd option for -loginhint, in addition to tel and ipport. This extension is only added for CIBA flow.
However, also authorization code flow would greatly benefit from the ability to pass the operator token. It shall be possible to leverage the -loginhint parameter in the /authorize request and lower the burden on Authorization server to convert an IP address to an MSISDN.

**Possible evolution**
Proposal is to allow loginhint parameter in the authorization code flow  authorization request and extend the description in the  CAMARA-Security-Interoperability.md (https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md) as follows:

Authorization code flow allows the client to use id_token_hint or login_hint as a hint in the authentication request. The client shall be enabled to use the optional login_hint. If used, the login_hint must be populated with one of the 3 formats described in "Format of login_hint" section (https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#format-of-login_hint)

**Alternative solution**

**Additional context**

AxelNennker:
The new format was added in https://github.com/camaraproject/IdentityAndConsentManagement/pull/218
The format section applies to all flows.
So, operator token support is already part of CAMARA. CIBA and OIDC authorization code flow.

jpengar:
> The new format was added in #218 The format section applies to all flows. So, operator token support is already part of CAMARA. CIBA and OIDC authorization code flow.

I disagree as mentioned here: https://github.com/camaraproject/IdentityAndConsentManagement/issues/191#issuecomment-2301498137 and here https://github.com/camaraproject/IdentityAndConsentManagement/issues/215#issuecomment-2470478556

>This issue was originally intended to clarify the authentication method for the Auth Code flow as defined by CAMARA today. I don't see operator tokens in the scope of this issue discussion.
>
[...]

PR #238, opened 2024-11-26
Add a section on operator token usage in authorization code flow

AxelNennker:
#### What type of PR is this?

* enhancement/feature

Provide a guideline how an operator token can be used.

### DIFF
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -247,7 +247,9 @@ How the Operator Token is retrieved from the API provider is out of scope of thi
 
 The format of the Operator Token is out of scope fo this document. If the Operator Token contains personal information that confidential information must be protected. If the Operator Token contains routing information that e.g. the Aggregator needs to know then that routing information can be encrypted to the Aggregator and the personal information of the User can be encrytpted to the API Exposure platform.
 
-Operator Token is standardized be GSMA in [TS.43](https://www.gsma.com/newsroom/gsma_resources/ts-43-service-entitlement-configuration/) and [ASAC](https://www.gsma.com/newsroom/gsma_resources/asac-01/).
+The Operator Token that is send to the Operator's Authorization Server is the token that was created by the Operator's Entitlement Server. If the Aggregator received a token with routing information then the Aggregator removes that routing information and sends the original Entitlement Server created token to the Operator's Authorization Server.
+
+Note: the value that follows the "operatortoken:" prefix is not nessesarily the operator token as defined in GSMA TS.43 v12 section 14 "Device App authentication". Some implementations send a GSMA TS.43 temporary token to the Operator's Authorization Server.
 
 The API Consumer (for example, the Application Backend) instructs the Application on the Consumption Device to initiate the OIDC Authorization Code Flow with the Operator. The authorization request includes the client_id of the ASP's Application requesting access to the API and the Application's redirect_uri (api_consumer_callback) where the authorization code will be sent. The authorization request includes the operator token in the `login_hint` OIDC parameter.
 


jpengar (commented):
I believe this is related to Issue #232 and/or an attempt to resolve it, isn't it?. However, as stated in https://github.com/camaraproject/IdentityAndConsentManagement/issues/232#issuecomment-2493649776, Telefónica has a different view on this proposal. For further details, please refer to the linked comment.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -184,6 +185,115 @@ The Application on the Consumption Device must be able to handle browser redirec
...
+
+The API Exposure Platform receives the request from the Application (Step 3) and does the following:
+
+- Decrypt and validate the operator token. Obtain the Subscriber's unique identifier,e.g.: phone number or IMSI. Sets the id_token sub to a unique user ID and associates the sub with the access token. The id_token sub MUST NOT reveal information to the Application, as authentication has not yet been performed. (Step 4).
+
+- Checks if User Consent is required, which depends on the legal basis associated with the Scope and Purpose. If necessary, it will check in the API Provider's Consent Master whether User Consent has already been given for this Application, Scope and Purpose (Steps 5-6).
+
+Then, two alternatives may occur:
+
+**Scenario 1**: User Consent is not required or User Consent has already been given (Step 7). The API Exposure Platform will continue the authorization code flow by redirecting to the Application's redirect_uri (api_consumer_callback) and including the authorization code (OperatorCode).
```
eric-murray, line 256: So, to be clear:
- if the `/authorize` request includes the operator token in a `login_hint`, and the User associated with that operator token has already given consent for the requested scope and purposes
- then the API transaction would proceed with **no further interaction with the end user of the device at all?**

I had understood that the purpose of mandating support for `login_hint` in the authorisation code flow was to allow "inband" interaction with the user of the device to get End User consent, using the `login_hint` only as a "hint" as to the true identity of the End User.

So why is Scenario 1 even an option? Surely this End User interaction should be mandatory when `login_hint` is included in the authorisation code flow, as their identity will not be known until they do interact.

sebdewet: As I understand it, adding the operator_token in the login_hint is the equivalent of a network authentication, the operator_token is a proof of authentication of the device (or identification of the SIM card) and it means that as a network authentication, if the consent is already given for this MSISDN (and not revoked), we don't need to wake up the end-user to get it.

jpengar: @sebdewet From the CAMARA perspective, the Auth Server receives an /authorize request with a login_hint containing an operator token. What happens if the app developer uses the operator token obtained for the SIM in device A (MSISDN A) in an Auth code flow from device B with a different SIM (MSISDN B)? If consent is granted or not required for MSISDN A, then the Auth server would issue an access token associated with MSISDN A, which is different from the consuming device B. Now, if the app in device B consumes, say, Device Location API, you could end up getting the location of device A instead of B. Or what happens if the operator token is used by another third party? We can agree that an operator token is much harder to misuse than an MSISDN, but provided as a login_hint (without further validation), from a CAMARA perspective it only identifies a device as an IP or an MSISDN when provided as a login_hint.

jpengar: In the last ICM call [04/12](https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/49971205/2024-12-04+ICM+Minutes) it was proposed that tokens be enhanced with attributes such as app IDs or IP addresses in order to improve security.

sebdewet: Why an app developper would do that ?
The application developer wants to provide and sell a service, if they misuse the operator token, they will not gain from the offered service.
There are certainly cases where this could be interesting for them, and we need to protect against that. In this case, we should think about identifying those cases and only offer the login_hint with the operator token in use cases where we are certain that misuse cannot benefit a developer.

PR #244, opened 2025-01-14
Allow API to decide on network-based authentication in OIDC authorization code flow

AxelNennker:
#### What type of PR is this?

* enhancement/feature

Currently CAMARA APIs have no way to specify in the [OIDC authentication request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) that the API requests network-based authentication or not. 

The OIDC standard allows the API consumer to specify the parameter `acr_values` to request [Authentication Context Class Reference values](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest:~:text=OPTIONAL.%20Requested-,Authentication%20Context%20Class%20Reference,-values.%20Space%2Dseparated).
The OIDC standard does not standardize possible values for the `acr_values` parameter.
IETF standardized some values for Authentication Method Reference Values, which are related to Authentication Context Class Reference values, in [RFC8176](https://datatracker.ietf.org/doc/html/rfc8176) but none of the standardized values match network-based authentication.

The PR defines a value https://camaraproject.org/acr/networkbasedauthentication for the OIDC acr_values parameter.

##### Example CAMARA APIs that require networkbased-authentication:
NumberVerification currently specifies in its yaml-file that it needs networkbased-authentication. 
No other API currently specifies the need for networkbased-authentication.
After this PR is merged the NumberVerification API specification can require that the API consumer sends Authentication Requests with the parameter acr_values="https://camaraproject.org/acr/networkbasedauthentication"

If the OIDC authentication request starts on the user's mobile phone, either in the API consumer's mobile application or in a browser, and the connection is "on-net", then a request with acr_values="https://camaraproject.org/acr/networkbasedauthentication" triggers network-based authentication which succeeds, regardless whether e.g. prompt=none or prompt=consent. The subscription is then identified and the flow continues with a consent-check for that subscription.

##### Benefits if the PR is merged and implemented

This PR allows the API provider to request network-based authentication.
This PR specifies one acr_values value https://camaraproject.org/acr/networkbasedauthentication. That value is also used in the returned ID Token als `amr_values` indicating whether network-based authentication was done or not.
The implementation of the authorization server's authorization code flow becomes more flexible.
Current implementations of OIDC authorization code flow become more standard.

### DIFF
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -100,7 +100,7 @@ Note over FE,BE: Use Feature needing<br>Operator Capability
 BE->>FE: Auth Needed - redirect <br>/authorize?response_type=code&client_id=coolApp<br>&scope=dpv:<purposeDpvValue> scope1 ... scopeN<br>&redirect_uri=invoker_callback...
 FE->>+FE: Browser /<br> Embedded Browser
 alt Standard OIDC Auth Code Flow between Invoker and API Exposure Platform
-  FE-->>ExpO: GET /authorize?response_type=code&client_id=coolApp<br>&scope=dpv:<purposeDpvValue> scope1 ... scopeN<br>&redirect_uri=invoker_callback...
+  FE-->>ExpO: GET /authorize?response_type=code&client_id=coolApp<br>&scope=dpv:<purposeDpvValue> scope1 ... scopeN<br>&acr_values=https%3A%2F%2Fcamaraproject.org%2Facr%2Fnetworkbasedauthentication<br>&redirect_uri=invoker_callback...
   Note over ExpO: API Exposure Platform applies<br>Network Based Authentication (amr=nba/mnba)
   ExpO->>ExpO: Network Based Authentication:<br>- map to Telco Identifier e.g.: phone_number<br>- Set UserId (sub)  
   ExpO->>ExpO: Check legal basis of the purpose<br> e.g.: contract, legitimate_interest, consent, etc 
@@ -136,7 +136,7 @@ As per the standard authorization code flow, the Application is redirected to th
 
 The Operator's API Exposure Platform receives the request from the Application (Step 3) and does the following:
 
-- Uses network based authentication to obtain the Subscriber's unique identifier,e.g.: phone number or IMSI. Sets the id_token sub to a unique user ID and associates the sub with the access token. The id_token sub MUST NOT reveal information to the Application, as authentication has not yet been performed. (Step 4).
+- If the parameter `acr_values=https%3A%2F%2Fcamaraproject.org%2Facr%2Fnetworkbasedauthentication` is received, then the authorization server uses network based authentication to obtain the Subscriber's unique identifier,e.g.: phone number or IMSI. Sets the id_token sub to a unique user ID and associates the sub with the access token. The id_token sub MUST NOT reveal information to the Application, as authentication has not yet been performed. (Step 4). If network-based authentication is not requested, then  the authorization server continues with the flow as defined in the OIDC standard.
 
 - Checks if User Consent is required, which depends on the legal basis associated with the Scope and Purpose. If necessary, it will check in the Operator's Consent Master whether User Consent has already been given for this Application, Scope and Purpose (Steps 5-6).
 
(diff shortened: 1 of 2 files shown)


AxelNennker:
For backward compability I added text to the PR that states that network-based authentication is the default.
Please review

jpengar:
I see https://github.com/camaraproject/IdentityAndConsentManagement/pull/244/commits/1298eb442a66aacb08094fcb11f2aaab03cdf039 considers the backwards compatibility which is good. Network-based auth should really be the default behaviour to ensure backwards compatibility (and also to ensure that you only "annoy" the user or interrupt the UX when absolutely necessary). But Telefónica still have concerns about the proposed solution:

- It only distinguishes between network-based auth and non-network-based auth, where all other methods would be included.
- It doesn't solve the inconsistency in UX or developer experience depending on the operator the user belongs to, if each operator can use the non-network-based auth method they want, user/pass, SMS OTP, others without further specification.
- It doesn't solve how to guarantee that there is always a token associated with a device/phone number, as CAMARA currently requires to delegate to the 3-legged token when accessing the APIs. This can be a problem for example in the user/password auth scenario.
- We think that consideration should be given to standardising each auth method to be enabled in CAMARA, having an acr value for each and checking the implications of each, which can be different. So, for example, if a developer asks to use user/pass, it is because they explicitly know that the UX will be interrupted and they will have considered this in the experience of their application. And you can have a deterministic and consistent experience. 
  - And if an operator doesn't support a particular acr, then CAMARA can define the corresponding error scenario and it could be standardised.
- acr_values suggested looks very long, we can consider other possible values to simplify the authentication request.

AxelNennker:
> I see [1298eb4](https://github.com/camaraproject/IdentityAndConsentManagement/commit/1298eb442a66aacb08094fcb11f2aaab03cdf039) is concerned about the backwards compatibility, which is good. Network-based auth should really be the default behaviour to ensure backwards compatibility (and also to ensure that you only "annoy" the user or interrupt the UX when absolutely necessary). But Telefónica still have concerns about the proposed solution:
> 
>     * It only distinguishes between network-based auth and non-network-based auth, where all other methods would be included.

The PR is looking for the minimal step forward that expands the current limits.

>     * It doesn't solve the inconsistency in UX or developer experience depending on the operator the user belongs to, if each operator can use the non-network-based auth method they want, user/pass, SMS OTP, others... without further specification.

These inconsistencies cannot be resolved, because that would force all operators to implement a common end-user authentication method. But users are used to authenticate to their operator with the same method they always use with that operator. Forcing all operators to a common (lowest) denominator would be confusing for end-users.
Even with one operator different users can configure their login-experience depending on their security needs.
E.g. Telekom Deutschland GmbH supports passkeys if users want them
CAMARA should stay away from end-user authentication methods and let the operator implement what is best for their users and security needs.
If CAMARA wants to go there, that would be another PR, I think.

>     * It doesn't solve how to guarantee that there is always a token associated with a device/phone number, as CAMARA currently requires to delegate to the 3-legged token when accessing the APIs. This can be a problem for example in the user/password auth scenario.

Network-based authentication assumes that there is one subscriber with one account. Family accounts and business accounts are a problem because in those cases the end-user is not necessarily the subscriber. E.g.: An employee uses their company-phone, then the subscriber is the employer and the end-user/employee might be allowed to grant consent or opt-out or whatnot. 
The same is true, if the end-user uses username(email) and password. Some end-users have one subscription, then the access token can be associated with a subscription identifier (MSISDN, IMSI, ...), while some end-users have multiple-subscriptions.

CAMARA currently does not define what happens in these cases. We could define a common behavior and probably should do that, but this problem is a problem for another PR in which e.g. login_hint is used to "suggest" the Target Device while the user still used e.g. their passkey to login and then grant consent.
If the end-user has 4 contracts, and login_hint suggest one of them we could recommend to operators to use that one.
Again, that is another PR.

> 
>     * We think that consideration should be given to standardising each auth method to be enabled in CAMARA, having an acr value for each and checking the implications of each, which can be different. So, for example, if a developer asks to use user/pass, it is because they explicitly know that the UX will be interrupted and they will have considered this in the experience of their application. And you can have a deterministic and consistent experience.

As I said above, this PR is a minimal step forward to expands the limits, and to let APIs and API Consumer chose what they need or expect from an API. 
The API consumer can always use prompt=none to check for consent_required.
If consent_required, then the developer can chose an appropriate and convenient point in time, when to use prompt=consent, or max_age=0 to start a end-user interaction which can't avoided anyway if consent is required.
As I said above, I think that the operator choses the end-user authentication method that their end-users expect and are used to. CAMARA should not enforce a lowest denominator. Operators are never going to agree on a common end-user authentication method. Even in one operator product owner often do not agree on login UI.
There might be an ADULT-PIN for TV streaming login, but passkey for contract management, and username password for "low-risk" operations (whatever that is)

>       
>       * And if an operator doesn't support a particular acr, then CAMARA can define the corresponding error scenario and it could be standardised.

Could be, but I would not do that. See above

>     * acr_values suggested looks very long, we can consider other possible values to simplify the authentication request.

I do not care much. We can have some bikeshedding session to find better values e.g. urn:nba and urn:nonba

Thanks, for your thoughts and comment.

PR #249, opened 2025-01-20
Revert "recommend auth code flow using signed requests"

AxelNennker:
Reverts camaraproject/IdentityAndConsentManagement#226

@jpengar wrote:
> I believe this PR (https://github.com/camaraproject/IdentityAndConsentManagement/pull/226) was merged prematurely, as key concerns raised during the discussion remain unresolved. Specifically:
>
>    The choice between OIDC and RFC9101 and the rationale for deviating from the latter.
>    The handling of the aud parameter, which conflicts with existing specifications like CIBA.
>    Questions about how to treat parameters not included in the signed request object, as RFC9101 specifies they must be ignored.
>
>Since these points were still under discussion and the proposed text was not agreed upon by all participants, I recommend reverting this PR until a consensus is reached and the final text is approved.

@AxelNennker @mhfoo @garciasolero @sebdewet @eric-murray @shilpa-padgaonkar

My mistake, please rest assured that I merged with good intentions. I thought all issues were addressed when I looked at the PR's diff page and saw no change requests to the PR.

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -9,7 +9,6 @@
     + [Transport Security](#transport-security)
     + [Sender-Constrained Tokens](#sender-constrained-tokens)
   * [OIDC Authorization Code Flow](#oidc-authorization-code-flow)
-    + [Signed Authentication Requests](#signed-authentication-requests)
     + [Optional Parameters](#optional-parameters)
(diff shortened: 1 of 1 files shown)


jpengar (approved):
LGTM. Thanks Axel. I guess we will need to open another analogous PR to #226 in order to reach consensus and get the final text approved.

PR #253, opened 2025-01-21
Agreed conclusion statement about auth method in Auth Code Flow

### LINKED ISSUE #215: Authentication method in authorization code flow

jpengar:
#### What type of PR is this?

* documentation

This PR includes a statement explaining the conclusion reached in #215 in the profile document as part of the Spring25 meta-release to make it 100% clear. As agreed [here](https://github.com/camaraproject/IdentityAndConsentManagement/issues/215#issuecomment-2595645711).

Fixes #215

N/A

#### Changelog input

```
 Added a concluding statement about the authentication method required in the Auth Code flow.
```

#### Additional documentation 

N/A

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -88,9 +88,9 @@ The following table defines the REQUIRED behaviour of the API Provider for the `
 
 ## OIDC Authorization Code Flow
 
-The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html). CAMARA REQUIRES API providers to use network-based authentication in the Authorization Code Flow. Therefore, access tokens are issued for the network authenticated identifier.
+The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html). CAMARA REQUIRES API Providers to use network-based authentication in the Authorization Code Flow. Therefore, access tokens are issued for the network authenticated identifier.
 
-It is up to the API Provider to make a decision when consent collection is needed or not, based on the scope(s)/purpose declared by the API consumer and aligned with local legislation, ensuring that all operators under the same regulatory framework adopt a consistent approach. If consent is required, the API provider SHALL obtain it according to [OIDC section 3.1.2.4](https://openid.net/specs/openid-connect-core-1_0.html#Consent). CAMARA does not prevent the API Provider from using additional authentication methods as part of the in-band consent capture process if consent is required.
+It is up to the API Provider to make a decision when consent collection is needed or not, based on the scope(s)/purpose declared by the API Consumer and aligned with local legislation, ensuring that all API Providers under the same regulatory framework adopt a consistent approach. If consent is required, the API Provider SHALL obtain it according to [OIDC section 3.1.2.4](https://openid.net/specs/openid-connect-core-1_0.html#Consent). CAMARA does not prevent the API Provider from using additional authentication methods as part of the in-band consent capture process if consent is required.
 
 ---
 **NOTE**
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -88,7 +88,16 @@ The following table defines the REQUIRED behaviour of the API Provider for the `
 
 ## OIDC Authorization Code Flow
 
-The OIDC Authorization Code Flow is defined in [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html)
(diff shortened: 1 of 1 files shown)


eric-murray (approved):

murthygorty (commented):
Looks great to me too, thank you!

AxelNennker:
A question to clarify:

After this is merged the following request is still valid, right? Note the "prompt=consent".

```
GET /authorize?
    prompt=consent
    &response_type=code
    &scope=openid%20dpv%3AFraudPreventionAndDetection%20sim-swap%3Acheck%20sim-swap%3Aretrieve-date
    &client_id=s6BhdRkqt3
    &state=af0ifjsldkj
    &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb HTTP/1.1
Host: server.example.com
```

AxelNennker:
> @AxelNennker Yes, that request would still be valid.
> 
> But the API provider could still ignore the request to prompt for in-band end user consent if the legal basis for processing the request did not require it. The API consumer cannot force the API provider to seek end user consent when that is not required. Prompting a fraudster to approve a SIM Swap check would defeat the purpose of that API.

I don't think consent collection started by a request with "prompt=consent" is in-band.

You are right regarding sim-swap, my mistake. I copied the example from our examples document and added prompt=consent.

We have this template text:
> In cases where personal data is processed by the API and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of three-legged access tokens is mandatory. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.

I assume, that in some legislation for sim-swap the end-user probably does not have a right to opt-out. I agree that opting-out somewhat defeats the purpose of the sim-swap API but API Providers have to follow the law and if the law grants the users the right to opt-out then we have to implement that. Whether the API makes sense is our problem.
I think for numberverification users have the right to opt-out in some legislation.
Or for some other API, which we sell with some legitimate-use purpose, but still the users have the right to opt-out.
Or, regardless, of API, the users might have the right to opt-out.

> In cases where personal data is processed by the API and users can exercise their rights to grant consent or revoke consent, the use of three-legged access tokens is mandatory and the API consumer can trigger end-user consent management by using the request parameter `consent` with the value `consent`. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.

Example with fictional API where users can exercise their rights:
```
GET /authorize?
    prompt=consent
    &response_type=code
    &scope=openid%20dpv%3AAccountManagement%20privacy-respecting-api
    &client_id=s6BhdRkqt3
    &state=af0ifjsldkj
    &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb HTTP/1.1
Host: server.example.com
```

AxelNennker (changes requested):
use API provider instead of operator

jpengar:
This PR should be a mere formality, as the agreement on the issue itself has already been concluded. The purpose of this PR is to implement the conclusion that has already been reached. The PR does not mention the 'prompt' parameter, nor does it define anything beyond it. @shilpa-padgaonkar Could you please review?

eric-murray:
> I don't think consent collection started by a request with "prompt=consent" is in-band.

The proposed text uses the term "in-band", but without definition. I had assumed "in-band" meant the mechanism specified [here](https://github.com/camaraproject/IdentityAndConsentManagement/blob/r2.1/documentation/CAMARA-API-access-and-user-consent.md#authorization-code-flow-frontend-flow). If that is not correct, "in-band" should be defined, or replaced with a reference to that diagram.

Regarding purposes for which the legal basis is "opt-out" consent, if an end user has not opted-out, then they have consented, and no consent confirmation is required for the API execution to proceed. They must (somehow) be made aware that they have the right to opt-out, and there must be a mechanism to allow that, but that does not have to be a pop-up on their phone saying "You have the right to opt-out from this application requesting your personal information. Would you like to opt-out now?".

murthygorty:
A little broadly, as a tenet, I'm thinking..
- whenever a client makes a request to a provider be it OIDC or CIBA, some of the metadata are requests/suggestions. API Provider will use their legal basis and internal requirements to drive their end-user experience.
I realize it is not a specific statement, but it might help us in looking at fields like prompt, login_hint etc. no?

shilpa-padgaonkar (approved):
Based on the [comment](https://github.com/camaraproject/IdentityAndConsentManagement/pull/253#issuecomment-2607123107) from @jpengar that this text does not influence the prompt parameter, changes in this PR LGTM.

AxelNennker (approved):

PR #256, opened 2025-01-31
Document when FE flow is applicable with regards to involved devices

### LINKED ISSUE #255: Clarification on which device is authenticated in authorization code flow
**Problem description**
The CAMARA -API-access-and-user-consent document describes the authorization code flow as one option for API invocation.
It lacks to clarify that the authorization code flow can only be applied if the device, which is origin of the authorization code request, is also the target device of the API call, for which the authorization is requested.

It should be explicitly mentioned, that the sub claim, contained in the access token generated by the authorization server indicates the target device of the upcoming API call and the data scopes reflect the permissions (or consent) granted by the legal responsible party of the target device. 
It is highly recommended that an API implementation compares the access token content against the potential payload parameters indicating the target device.

**Expected action**
Update chapter on User Authentication/Authorization and Consent Management [(https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#user-authenticationauthorization--consent-management)] with corresponding clarification.
A pull request will be opened.

**Additional context**

Elisabeth-Ericsson:
#### What type of PR is this?
* documentation

There is some misunderstanding about Authorization code flow, a.k.a. FE flow.
It is important to clarify that OIDC authorization code flow is meant to authorize the client/device, which originates the authorization code flow request. This is independent of whether an optional login_hint (e.g. an operator token) is provided or not. The access code generated for such a request (and the resulting access token) are ONLY applicable to API calls, where the target device of the API call is the same device, which has sent in the authorize request earlier.
Authorization code flow must not be abused by indicating an identifier in the optional login_hint, which deviates from the device, which starts the flow.

Fixes #255 

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```


documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -58,6 +58,7 @@ The list below introduces several key concepts:
 -	`Resource Server`: the server that exposes protected resources to Applications. The Resource Server requires a valid access token to be provided before allowing access to the protected resource.
 -	`Scope`: the OpenID Connect scope which maps one or more protected resources, some scopes may require processing of Personal Data.
 - `Subscriber`: the mobile subscriber of the Operator. The Subscriber is usually also the End-User, but this is not always the case. For example, a parent may be the Subscriber of a mobile subscription for their child, the End-User.
+- `Target Device`: the device, which is the main resource (target) impacted by an API call and for which consent might need to be collected from resource owner.
```
jpengar, line 61: To me the `Target Device` is the `Consumption Device` which is already defined in the document. And the access token will be associated the network authenticated identifier corresponding to the `Consumption Device`. It has been clarified as well [in the profile](https://github.com/camaraproject/IdentityAndConsentManagement/blob/46beef349cef154e46e68e500cc05b36f9184c04/documentation/CAMARA-Security-Interoperability.md?plain=1#L102) `...CAMARA REQUIRES API Providers to use network-based authentication in the Authorization Code Flow. Therefore, access tokens are issued for the network authenticated identifier.`

So, IMHO, you should refer to the `Consumption Device` instead. You may need somehow to make your point about the differences between Consumption and Target device clearer.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -282,12 +286,13 @@ If some use case(s) for an API point to "Off-net" scenarios and where Consumptio
...
       - If necessary, it will be checked **in the operator's consent master** whether user consent has already been given to the application for the user identifier and declared purpose.
       - If NOT granted, **the operator’s consent capture procedure is triggered**. Out-of-band consent capture as part of asynchronous CIBA flow (e.g. push notification with fallback to SMS, etc...). **Operators can choose the consent capture mechanism that best suits their capabilities, preferences and needs**.
-  - Covered scenarios:
+  - Covered scenarios: 
     - No front-end developer software in user device
     - Back-end services (e.g. bank BE anti-fraud validation using MSISDN).
     - Off-net scenarios (no mobile connection)
     - Device connected to WiFi
     - Device without UI (IoT)
+    - Note: Consumption device and target device of subsequent API call(s) can be different, access token is generated for the target device
```
jpengar, line 295: I think this is already clarified by line 278 `If some use case(s) for an API point to "Off-net" scenarios and where Consumption Device and authentication devices could be different, the Backend flow should be used.`. Isn't it?

documentation/CAMARA-API-access-and-user-consent.md
jpengar:
```suggestion
- `Target Device`: the device that is the primary resource (target) affected by an API call, and for which Consent may need to be obtained from the Resource Owner.
```

AxelNennker:
I hope that misunderstandings regarding OIDC Authorization Code Flow were removed by "[Agreed conclusion statement about auth method in Auth Code Flow #253](https://github.com/camaraproject/IdentityAndConsentManagement/pull/253)"
That PR added text to [CAMARA Security and Interoperability Profile](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md).

A quote from the conclusion:
> CAMARA does not intend to limit the Authorization Code Flow to just network-based authentication in the future. New workstreams will be initiated to define the necessary changes that will ensure the Authorization Code Flow can be interoperably used with both network-based and non-network-based authentication methods.

I created several PRs that try to remove that "limit" e.g. https://github.com/camaraproject/IdentityAndConsentManagement/pull/244

We created [CAMARA Security and Interoperability Profile](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md) to clarify and add new meaning to e.g. [OIDC](https://openid.net/specs/openid-connect-core-1_0.html) and e.g. "abused" scope to introduce "[purpose as a scope](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#purpose-as-a-scope)". @shilpa-padgaonkar proposal to introduce a new purpose parameter would have been a "cleaner" solution, but that was rejected because suppliers prefer implementing standards and are reluctant to add new parameters. If ICM decides that e.g. `login_hint` has more meaning in the future, without violating the standard, then that decision would follow the same reasoning as what we did regarding purpose.
We could define a new parameter "target_device" or use "login_hint". I propose to explore how far we can get with standard parameters before adding new parameters.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -84,6 +85,7 @@ This section describes the authorization flows that can be used to access CAMARA
 Note: In cases where Personal Data is processed by a CAMARA API, and Users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of Three-Legged Access Tokens is mandatory.
 
 #### Authorization Code Flow (Frontend Flow)
+The authorization code flow is only applicable if the consumption device, which initiates the process of obtaining a Three-Legged-Access token from the Authorization Server, is also the target device of the intented Network API call(s) .
```
jpengar, line 88:
```suggestion
The Authorization Code Flow is only applicable if the Consumption Device that initiates the process of obtaining a Three-Legged Access Token from the Authorization Server is also the Target Device of the intended CAMARA API call(s).
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -180,12 +182,14 @@ The Application on the Consumption Device must be able to handle browser redirec
     - Check if User Consent is required as the lawful basis associated with the declared Scope and Purpose. 
       - If necessary, it will be checked in the operator's consent master whether user consent has already been given to the application for the user identifier and declared purpose.
     - If NOT granted, the operator performs the consent capture. Since the authorization code grant involves the interaction with application front-end, consent can be captured directly from the user through the application browser.
-  - Covered scenarios:
+  - Covered scenarios: 
     - On-net (with mobile connection) & application front-end (with embedded browser)
     - Off-net scenarios using refresh_token, as long as there was a connection when the first access_token was requested.
+    - Note: Consumption device must be equal to target device of subsequent API call(s).
```
jpengar, line 188:
```suggestion
    - Note: The Consumption Device must be equal to the Target Device of subsequent API call(s).
```
This would be already clarified in line 88, wouldn't be?

documentation/CAMARA-API-access-and-user-consent.md
jpengar: By definition, in an Auth code flow, the consumption device is the same as the authentication device. And CIBA is exactly the flow to use when this is not the case.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -180,12 +182,14 @@ The Application on the Consumption Device must be able to handle browser redirec
...
 
 
 #### CIBA flow (Backend flow)
+The CIBA flow is applicable if the consumption device is equal to or different from the target device of the intented Network API call(s).
```
jpengar, line 192:
```suggestion
The CIBA flow is applicable if the Consumption Device is equal to or different from the Target Device of the intended CAMARA API call(s).
```
I find this a bit confusing. May not be necessary...

jpengar (commented):
@Elisabeth-Ericsson I guess this clarification is still being added under Spring25 scope, where network-based auth must be used in Auth Code Flow. Right?

AxelNennker:
The purpose of this PR is clarification what the two three-legged flows actually do regarding "devices", right?

As discussed in #215 OpenId Authorization Code Flow **currently** does network-based authentication.
We added clarifying text in this PR.
https://github.com/camaraproject/IdentityAndConsentManagement/pull/253/files
That documents the status quo.

We are currently working on TS.43 based and SIM-based (EAP-AKA) authentication. Which is going to change that status quo.

@Elisabeth-Ericsson  Do we need this PR and the new definition of Target Device?

Elisabeth-Ericsson:
@AxelNennker : I think that we need this PR even more, if we allow more methods for authentication.
- If we allow the operator_token as login hint at some point in time, then also here it must be understood that the operator token identifies the consumption device, which also acts as target resource.
- If you want to extend/modify this with your ideas on using FE flow for testing consent, and want to introduce another interpretation, than is it even more needed to have a clear understanding on what we have today and where we want to go.
- When I talk to my colleagues from E///, they really appreciate this clarification, because this is the basis for potential changes & extensions. 
- We are writing our documentation for a broad audience and I learned that it is NOT properly understood by everbody that we have this "limitation" or call it "scenario" in the FE flow today that consumption device == target resource of API call.

documentation/CAMARA-API-access-and-user-consent.md
Elisabeth-Ericsson: > To me the `Target Device` is the `Consumption Device` which is already defined in the document. And the access token will be associated the network authenticated identifier corresponding to the `Consumption Device`. It has been clarified as well [in the profile](https://github.com/camaraproject/IdentityAndConsentManagement/blob/46beef349cef154e46e68e500cc05b36f9184c04/documentation/CAMARA-Security-Interoperability.md?plain=1#L102) `...CAMARA REQUIRES API Providers to use network-based authentication in the Authorization Code Flow. Therefore, access tokens are issued for the network authenticated identifier.`
> 
> So, IMHO, you should refer to the `Consumption Device` instead.
> 
> You may need somehow to make your point about the differences between Consumption and Target device clearer.

@Jesus: I added the definition to the top of the document, applicable to FE (auth code) flow and CIBA. And it is exactly the point of the CR to state that in case of FE (auth code) flow we have consumption device == target device, but this is not necessarily the case in CIBA.

documentation/CAMARA-API-access-and-user-consent.md
Elisabeth-Ericsson: > I think this is already clarified by line 278 `If some use case(s) for an API point to "Off-net" scenarios and where Consumption Device and authentication devices could be different, the Backend flow should be used.`. Isn't it?

No, this is NOT my point, I'm after the statement that the consumption device (where the app runs) and the TARGET device of an API call might be different in case of CIBA. Use case: "Find My Pet".
The the consumption device and the auth device might be different as well.
We have 3 different roles ("devices"): Consumption device (where the app runs), target device/resource (target of the API call), Authorization device: contact information identifying the legal responsible of a target device

Elisabeth-Ericsson: > By definition, in an Auth code flow, the consumption device is the same as the authentication device. And CIBA is exactly the flow to use when this is not the case.

I do NOT fully agree.
In Auth code flow, the consumption device is the same as the target device.
But whether the consumption device might give consent or not is a different question, e.g. Kid playing a game. Then the consumption device and the target device of the API call, e.g. for device location is the Kid's device, but the auth device might be the parent's phone.

So, I agree that we are addressing a limited set of use cases in CAMARA today. But we need to clarify the roles of the involved stakeholders, end users and resource owners. For simplicity named "consumption device", "target device", "authorization device".

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -189,7 +189,7 @@ The Application on the Consumption Device must be able to handle browser redirec
 
 
 #### CIBA flow (Backend flow)
-The CIBA flow is applicable if the consumption device is equal to or different from the target device of the intented Network API call(s).
+The CIBA flow is applicable if the Consumption Device is equal to or different from the Target Device of the intended CAMARA API call(s).
```
AxelNennker, line 192: CIBA is necessary if Consumption Device and Authentication Device are different. If Consumption Device and Authentication Device are the same, then OIDC Authentication Code Flow SHOULD be used.

The API Provider MUST send a message to the authentication device, identified by login_hint, because otherwise there would be no authentication at all. 

Note: In cases where personal data is processed by the API and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of three-legged access tokens is mandatory. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.

User consent cannot be checked based on the login_hint value alone and without sending a message, because then CIBA would be a two-legged flow while according the the note above a three-legged flow is mandatory.

AxelNennker (changes requested):

documentation/CAMARA-API-access-and-user-consent.md
Elisabeth-Ericsson: **Relation between consumption device and authentication device:**
>CIBA is necessary if Consumption Device and Authentication Device are different. If Consumption Device and >Authentication Device are the same, then OIDC Authentication Code Flow SHOULD be used.

The API invoker cannot know whether consumption device and authentication device are different. The API invoker in general has no awareness at all about authentication devices. The API invoker only knows the target device and potentially also the consumption device. Who is the legal responsbile (authorization device) of the target device is only known to the backend - the telco operator.

The OIDC Authentication Code Flow protocol does not foresee to communicate whether the consumption device is the authentication device or not. It would be beneficial to get this information in the response, but this is not the case today.

**CIBA flow:**
>User consent cannot be checked based on the login_hint value alone and without sending a message, because then >CIBA would be a two-legged flow while according the the note above a three-legged flow is mandatory.
Can you please clarify this statement ? 
AFAIK CIBA transports the information about the target resource/target device in the login_hint, The consent check happens for the target resource specified in the login_hint, The API invoker does not have any knowlege about the consumption device in this context.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -180,12 +182,14 @@ The Application on the Consumption Device must be able to handle browser redirec
...
     - If NOT granted, the operator performs the consent capture. Since the authorization code grant involves the interaction with application front-end, consent can be captured directly from the user through the application browser.
-  - Covered scenarios:
+  - Covered scenarios: 
     - On-net (with mobile connection) & application front-end (with embedded browser)
     - Off-net scenarios using refresh_token, as long as there was a connection when the first access_token was requested.
+    - Note: The Consumption Device must be equal to the Target Device of subsequent API call(s).
 
 
 #### CIBA flow (Backend flow)
+The CIBA flow is applicable if the Consumption Device is equal to or different from the Target Device of the intended CAMARA API call(s).
```
jpengar, line 192: The applicability of the CIBA Flow is described in the `Technical ruleset for the Backend flow` section. It may be better to move this statement there and integrate it with the rest of the information in that section [here](https://github.com/camaraproject/IdentityAndConsentManagement/blob/bd65974fa496ec065c9da2acbda448e215c1bb9c/documentation/CAMARA-API-access-and-user-consent.md?plain=1#L272)

Something like...

```
If the API use case(s) indicate the need for an "Off-net" scenario and/or if the Consumption Device and Authentication Device are different, the Backend flow (CIBA) SHOULD be used.

Also note that in the case of the CIBA flow, it is applicable if the Consumption Device is the same or different from the Target Device of the intended CAMARA API call(s).
```

jpengar (commented):
@Elisabeth-Ericsson I've added a few more comments to ensure the document's overall meaning and structure. Please take a look and consider replacing existing content with my suggestions if you agree.

jpengar:
@Elisabeth-Ericsson Hi Eli, this PR and the corresponding issue (#255) were included in the scope of the 'Fall25' meta-release. However, there have been no updates for more than two months since my review. As discussed in [yesterday's ICM call](https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/123077145/2025-05-21+Draft+Agenda+ICM+Minutes), the WG is behind schedule in closing the necessary scope points (https://github.com/camaraproject/IdentityAndConsentManagement/milestone/1) in order to meet the M2 milestone by 7 June. Therefore, I suggest trying to update and close this PR by next week, or removing it from the scope of the next meta-release. CC @subha5h

documentation/CAMARA-API-access-and-user-consent.md
Elisabeth-Ericsson: agree, let's move

Elisabeth-Ericsson:
@jpengar , @AxelNennker I have included the proposed changes in the document after having updated the branch to the latest release.
Please review the latest changes.

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -171,9 +170,10 @@ Finally, the Operator will provide the API response to the Application (Step 15)
 
 _NOTE: The technical ruleset is applicable only after a subproject has agreed to use a Three-Legged Access Token authentication flow. This ruleset provides a recommendation which will help API providers to align on the Three-Legged Access Token Flow and help with aggregation._
 
-If all API usecases point to the need of an 'On-Net' scenario and where the Consumption Device and Authentication Device are the same, the Frontend flow SHOULD be used. eg. NumberVerification
+If all API usecases point to the need of an 'On-Net' scenario and where the Consumption Device and Authentication Device are the same, the Frontend flow SHOULD be used. eg. NumberVerification.
```
jpengar, line 173:
```suggestion
If all API use cases point to the need of an 'On-Net' scenario and where the Consumption Device and Authentication Device are the same, the Frontend flow SHOULD be used e.g. Number Verification. 
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -171,9 +170,10 @@ Finally, the Operator will provide the API response to the Application (Step 15)
...
+Also note, the Authorization Code Flow is only applicable if the Consumption Device, that initiates the process of obtaining a Three-Legged Access Token from the Authorization Server, is also the Target Device of the intended CAMARA API call(s).
 
-This flow is then applicable to On-Net scenarios where the mobile connection of the Consumption Device needs to be authenticated e.g. [CAMARA Number Verification API](https://github.com/camaraproject/NumberVerification) due to the nature of its functionality where a User's phone number needs to be compared to the phone number associated with the mobile connection of the Consumption Device. 
+This flow then enables On-Net scenarios where the mobile connection of the Consumption Device needs to be authenticated. for example, the [CAMARA Number Verification API](https://github.com/camaraproject/NumberVerification) due to the nature of its functionality where a User's phone number needs to be compared to the phone number associated with the mobile connection of the Consumption Device.
```
jpengar, line 176:
```suggestion
This flow then enables On-Net scenarios where the mobile connection of the Consumption Device needs to be authenticated. For example, the [CAMARA Number Verification API](https://github.com/camaraproject/NumberVerification) due to the nature of its functionality where a User's phone number needs to be compared to the phone number associated with the mobile connection of the Consumption Device. 
```

documentation/CAMARA-API-access-and-user-consent.md
```diff
@@ -51,6 +51,8 @@ The list below introduces several key concepts:
 - `API Exposure Platform`: the Operator's platform for exposing CAMARA APIs to ASPs and Aggregators, providing authentication and authorization mechanisms, and End-User Consent management. The API Exposure Platform typically consists of at least an Auth Server and an API Gateway.
 -	`Application` or `Application Backend`: the ASP's software services that access CAMARA APIs.
 - `Application Service Provider (ASP)`: the Legal Entity that provides the Application and/or services that consume CAMARA APIs.
+- `Authentication Device`: The device on which the user will authenticate/authorize the request, often a smartphone.
+Ref: https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#terminology
```
jpengar, line 55:
```suggestion
- `Authentication Device`: The device on which the user will authenticate/authorize the request, often a smartphone.
```

documentation/CAMARA-API-access-and-user-consent.md
jpengar: These are just typo corrections. I will commit the changes directly.

jpengar (approved):
LGTM

sebdewet (approved):
LGTM

AxelNennker (approved):

jpengar:
As agreed in today's ICM Fall25 Task Force session 2 meeting, I will proceed to merge this PR.

PR #262, opened 2025-02-06
Remove reference to formatting and encoding of operator token values

### LINKED ISSUE #259: Encoding of login_hint with operator token prefix
**Problem description**

When sending the `login_hint` with the operator token prefix in CIBA flow, the explanation of how its value should be handled if it contains non-url-safe characters is incorrect and may cause confusion. The [CIBA standard](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1) already specifies how such characters should be sent, and there is no need to encode them in base64.

**Expected action**

I suggest removing this sentence or changing it to something like this:

* **_operatortoken_**

    For operator tokens as defined by [GSMA TS.43](https://www.gsma.com/newsroom/gsma_resources/ts-43-service-entitlement-configuration/) and [GSMA ASAC](https://www.gsma.com/newsroom/gsma_resources/asac-01-v1-0/).
    TS.43 does not specify the format of the operator token and it therefore might contain characters that are not url-safe. Therefore, as with the other parameters, the API consumer MUST send the value in the `application/x-www-form-urlencoded` format and a character encoding of UTF-8.

**Additional context**

garciasolero:
#### What type of PR is this?

* documentation

It is not necessary to define a specific encoding format for the value of the operator token sent as login_hint in CIBA because it may contain url-unsafe characters. The standard already addresses this issue and states that all parameters must be encoded in `application/x-www-form-urlencoded`.

Any reference to the encoding of the operator token is removed, just as the other parameters that could contain url-unsafe characters are not mentioned for the same case, as this could lead to misunderstandings by sending a double-encoded parameter.

Fixes #259

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -175,10 +175,7 @@ This CAMARA document clarifies the values used in login_hint in the following wa
 
   * **_operatortoken_**
 
-    For operator tokens as defined by [GSMA TS.43](https://www.gsma.com/newsroom/gsma_resources/ts-43-service-entitlement-configuration/) and [GSMA ASAC](https://www.gsma.com/newsroom/gsma_resources/asac-01-v1-0/).
-    TS.43 does not specify the format of the operator token and it therefore might contain characters that are not url-safe.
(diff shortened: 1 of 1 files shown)


jpengar (approved):
LGTM

AxelNennker (approved):

Issue #224, opened 2024-11-07
Consent URL API vs OIDC consent collection


AxelNennker:
**Problem description**
Telefónica created a PR in API backlog regarding a new API named [Consent URL](https://github.com/camaraproject/APIBacklog/pull/67). 

It is DT's position that Consent URL API is not needed because at the same point in time Consent URL API would be used by the API consumer the API consumer can alternatively request an access token which also leads to consent being collected.

**Possible Outcome**

* Does ICM recommend creating a consent API? 
* Do we need something else than an API? 
* Do we need nothing?

**Additional Considerations**

* CAMARA identity and consent management is based on OIDC and CIBA. 
* The OIDC parameter `prompt` is [mandatory to implement by all OpenId providers](https://openid.net/specs/openid-connect-core-1_0.html#ServerMTI).
* Delegated Consent should/must be avoided.
* Consent collection MUST be controlled by the API provider.
* Consent collection in done according to [Section 3.1.2.4 of the OIDC Core 1.0 spec](https://openid.net/specs/openid-connect-core-1_0.html#Consent)
* Regarding consent collection [CIBA](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.8) also refers to Section 3.1.2.4 of the OIDC Core 1.0 spec.
* It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.

HuubAppelboom:
@AxelNennker What is missing in above list is the mangement of the consent, for which the API provider is responsible.
This should consist of showing the end user what consent has been given for which API consumer, at which date, for which purpose, including the exact details of the consent.  The API provider is also responsible for providing the user an option to withdraw the consent, and this should be as easy as how the consent was collected. As soon as the consent is withdrawn, all processing of the personal data should cease immediately.

In the ideal case, the end user should only give once consent for a case, and the text presented to the end user should inform him about both the processing that the API Provider will do, and the processing the API Consumer will do. It will also need to specify the purpose, so the end user can make an informed decision.

If you combine the above with the current specification, there are several issues that can occur. To name a couple:

[...]

Issue #258, opened 2025-02-03
CIBA Privacy


AxelNennker:
**Problem description**
An attacker, e.g misbehaving API Consumer or misbehaving Aggregator or somebody who stole the API Consumer's credentials, can send CIBA requests with arbitrary login_hints.
[CAMARA APIs Access and User Consent Management](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md) "validates the user identifier" in step 2 and then in step 3 checks the legal basis for the purpose and then in step 4 checks "if consent is granted".

Now the attack: if the AZ responds immediately responds then the attacker knows that the User did either granted consent or did not opt-out for this combination of request parameters - consent is not needed. So, the AZ is revealing personal data about the User (whether they consented, opted-out, ...)

The attacker with valid Credentials can create CIBA Authentication Requests and iterate through many MSISDNs. 

Let's say the attacker bought some MSISDNs from a databreach and wants to prepare an attack. The attacker can now check the consent-needed-status for an API and then either trick the users into opting-out of that API or focus their attack on Users who opted-out of some API, because attacks on Users who opted-out are more likely to succeed.

In the example attack the attacker does not (immediately) get an access token for an API, but the attacker gains personal data(the fact that he has opted out) about the User. We are violating the User's privacy.
 
Variant of this privacy-violating attack: If the CIBA Authentication Request is not signed, the Aggregator can manipulate the login_hint value and get personal data about a different User Identifier (MSISDN). The MNO has no protection against a misbehaving Aggregator.

OIDC Authorization Code Flow and CIBA add User authentication to OAuth2, but in the current CAMARA CIBA flow there is no User authentication if consent is not needed. The CIBA flow as shown in [CAMARA APIs Access and User Consent Management](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md) seems to be more like a two-legged flow because the current CIBA flow does not involve any User authentication.

![Image](https://github.com/user-attachments/assets/b06e854c-f415-42e7-869e-b0848ba40c95)

MarkCornall:
would you not also need to have stolen credentials to query the operator? Is that not limited to the scope of the API that access was previously granted access to for a nominated developer and application in the application onboarding?

HuubAppelboom:
[...]

Issue #49, opened 2023-07-11
What should be included as a part of a consent record?


shilpa-padgaonkar:
**Problem description**
To be able to design a consent API, we need to detail out what a consent record should include. This issue is created as a question to gather feedback from the subproject members.

Some examples could be: ConsentStatus (granted/denied), Consent capture date+time, PersonalData  etc.

**Possible evolution**
We should be able to define a clear structure for consent.

**Alternative solution**

**Additional context**

jpengar:
@shilpa-padgaonkar If it helps, and just as an example, below is how we currently model a consent record internally at Telefónica (not all fields may apply):

<img width="398" alt="image" src="https://github.com/camaraproject/IdentityAndConsentManagement/assets/1318242/6fd3d09b-f474-4330-a442-95274989eef9">

> NOTE: _We also want to add the application (with a client_id field or something like that)_.
> NOTE: _Fields like channel, reason, triggered_by... may be a bit more specific to Telefónica and not necessarily needed here_.

In any case, so far there is no need to expose a Consent API to third party clients as a CAMARA API with CRUD operations on consent records. Consent must be obtained through operator channels (there is no delegated consent) and only operators internally manage user consents (e.g. check if a consent record exists, create a new consent record if it is captured from the user, etc).

PR #266, opened 2025-02-11
Consent Info/Consent URL API DRAFT proposal

jpengar:
#### What type of PR is this?

* enhancement/feature

This PR is intended to provide a draft proposal for what the Consent URL or Consent Info API could look like (adding the functionality indicated by Ericsson/Vonage @Elisabeth-Ericsson @chrishowell).

**This PR is not intended to be merged, as if we were to proceed with the API specification in CAMARA, I understand that the API should go into its own repository to have its own release management and where the fine-grained details of the design could be discussed.**

This DRAFT proposal may help to resolve the API approval discussions in Issue #224. However, the proposal already contains a good level of detail and is aligned with the Sprin25 Commonalities and ICM guidelines (just for reference).

Is related to Issue #224 

* Importantly, this API does NOT delegate consent capture to a third party but rather empowers the API Consumer to present the API Provider's consent capture URL at the most opportune time and place. The actual consent capture occurs within the API Provider's secure environment, ensuring the user's authentication with the API Provider.
* This API adds a layer of flexibility to the consent collection process, allowing API Consumers to more seamlessly integrate and customise the consent experience, without intending to replace existing CAMARA OIDC/CIBA flows. Instead, it complements these flows by providing more options for when and how consent is captured.

#### Changelog input

```
consent-info draft proposal
```

#### Additional documentation 

https://github.com/camaraproject/IdentityAndConsentManagement/issues/224
https://github.com/camaraproject/APIBacklog/issues/68
https://github.com/camaraproject/APIBacklog/pull/67

### DIFF
--- a/documentation/API_documentation/consent-info.yaml
+++ b/documentation/API_documentation/consent-info.yaml
@@ -17,7 +17,9 @@ info:
 
     # API Functionality
 
-    This API enables an API Consumer to determine whether user Consent is required, in the applicable legislation, for the requested actions based on the provide scope(s) and Purpose. Specifically, the API:
+    This API enables an API Consumer to determine whether user Consent is required, in the applicable legislation, for the requested actions based on the provide scope(s) and Purpose. NOTE: It may be required more than one User Consent for the requested scope(s) and Purpose (e.g. when the requested scopes are related to multiple APIs).
+    
+    Specifically, the API:
 
     * Determines the need for Consent: Evaluates whether user Consent is necessary according to the specified parameters and relevant legal requirements.
     * Provides the current Consent status: If Consent is required, the API returns the current status (e.g., active, revoked, expired...) so the API Consumer can understand the current User's consent status.
@@ -34,6 +36,20 @@ info:
     The specific authorization flows to be used will be agreed upon during the onboarding process, happening between the API consumer and the API provider, taking into account the declared purpose for accessing the API, whilst also being subject to the prevailing legal framework dictated by local legislation.
 
     In cases where personal data is processed by the API and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of three-legged access tokens is mandatory. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.
+    
+    # Identifying the phone number from the access token
+
+    This API requires the API consumer to identify a phone number as the subject of the API as follows:
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` field, which therefore MUST be provided.
+    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
+
+    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
+
+    ## Error handling:
+
+    - If the subject cannot be identified from the access token and the optional `phoneNumber` field is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
+
+    - If the subject can be identified from the access token and the optional `phoneNumber` field is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same phone number is identified by these two methods, as the server is unable to make this comparison.
     
     # Further info and support
 
@@ -57,7 +73,7 @@ paths:
     post:
       summary: Create a request to verify the Consent status
       description: |
-        Create a request to verify the Consent status of a specific User to the API Consumer for the requested scope(s) and Purpose. The User is identified by the `phoneNumber` parameter in the request body. The API Consumer is identified by the `client_id` parameter deduced from the access token.
+        Create a request to verify the Consent status of a specific User to the API Consumer for the requested scope(s) and Purpose. The API Consumer is identified by the `client_id` parameter deduced from the access token.
       operationId: verifyConsentStatus
       security:
         - openId:
@@ -74,12 +90,42 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/VerifyConsentStatusRequestBody"
-            example:
-              phoneNumber: "+123456789"
-              scopes:
-                - "location-verification:verify"
-              purpose: "dpv:FraudPreventionAndDetection"
-              requestCaptureUrl: true
+            examples:
+              ONE_SCOPE_ONE_API:
+                summary: One scope
+                description: |
+                  Request with one scope corresponding to one API
+                value:
+                  phoneNumber: "+123456789"
+                  scopes:
+                    - "location-verification:verify"
+                  purpose: "dpv:FraudPreventionAndDetection"
+                  requestCaptureUrl: true
+              MULTIPLE_SCOPES_ONE_API:
+                summary: Multiple scopes for one API
+                description: |
+                  Request with multiple scopes corresponding to one API
+                value:
+                  phoneNumber: "+123456789"
+                  scopes:
+                    - "quality-on-demand:sessions:create"
+                    - "quality-on-demand:sessions:read"
+                    - "quality-on-demand:sessions:update"
+                    - "quality-on-demand:sessions:delete"
+                    - "quality-on-demand:sessions:retrieve-by-device"
+                  purpose: "dpv:RequestedServiceProvision"
+                  requestCaptureUrl: true
+              MULTIPLE_SCOPES_MULTIPLE_APIS:
+                summary: Multiple scopes for multiple APIs
+                description: |
+                  Request with multiple scopes corresponding to multiple APIs
+                value:
+                  phoneNumber: "+123456789"
+                  scopes:
+                    - "location-verification:verify"
+                    - "device-roaming-status:read"
+                  purpose: "dpv:FraudPreventionAndDetection"
+                  requestCaptureUrl: true
       responses:
         "200":
           description: OK
@@ -96,21 +142,66 @@ paths:
                   description: |
                     Consent is not required for the requested scope(s) and Purpose
                   value:
-                    consentRequired: false
+                    privacyStatus:
+                      - scopes:
+                          - "number-verification:verify"
+                        purpose: "dpv:LegitimateInterest"
+                        consentRequired: false
                 CONSENT_REQUIRED_ACTIVE:
                   summary: Consent is required and active
                   description: |
                     Consent is required for the requested scope(s) and Purpose and it is provided by the User
                   value:
-                    consentRequired: true
-                    consentStatus: ACTIVE
+                    privacyStatus:
+                      - scopes:
+                          - "location-verification:verify"
+                        purpose: "dpv:FraudPreventionAndDetection"
+                        consentRequired: true
+                        consentStatus: ACTIVE
                 CONSENT_REQUIRED_CAPTURE_URL:
                   summary: Consent is required and capture URL provided
                   description: |
                     Consent is required for the requested scope(s) and Purpose but it is not provided by the User.
                   value:
-                    consentRequired: true
-                    consentStatus: PENDING
+                    privacyStatus:
+                      - scopes:
+                          - "location-verification:verify"
+                        purpose: "dpv:FraudPreventionAndDetection"
+                        consentRequired: true
+                        consentStatus: PENDING
+                    captureUrl: 'https://example.org/consent-capture-url'
+                CONSENT_REQUIRED_MULTIPLE_SCOPES_ONE_API:
+                  summary: Multiple scopes for one API
+                  description: |
+                    Consent is required for the requested scope(s) and Purpose but it is not provided by the User. Request with multiple scopes corresponding to one API. 
+                  value:
+                    privacyStatus:
+                      - scopes:
+                        - "quality-on-demand:sessions:create"
+                        - "quality-on-demand:sessions:read"
+                        - "quality-on-demand:sessions:update"
+                        - "quality-on-demand:sessions:delete"
+                        - "quality-on-demand:sessions:retrieve-by-device"
+                        purpose: "dpv:RequestedServiceProvision"
+                        consentRequired: true
+                        consentStatus: PENDING
+                    captureUrl: 'https://example.org/consent-capture-url'
+                CONSENT_REQUIRED_MULTIPLE_SCOPES_MULTIPLE_APIS:
+                  summary: Multiple scopes for multiple APIs
+                  description: |
+                    Consent is required for the requested scope(s) and Purpose but it is not provided by the User. Request with multiple scopes corresponding to multiple APIs.
+                  value:
+                    privacyStatus:
+                      - scopes:
+                          - "location-verification:verify"
+                        purpose: "dpv:FraudPreventionAndDetection"
+                        consentRequired: true
+                        consentStatus: PENDING
+                      - scopes:
+                          - "device-roaming-status:read"
+                        purpose: "dpv:FraudPreventionAndDetection"
+                        consentRequired: true
+                        consentStatus: PENDING
                     captureUrl: 'https://example.org/consent-capture-url'
         "400":
           $ref: "#/components/responses/Generic400"
@@ -147,7 +238,6 @@ components:
     VerifyConsentStatusRequestBody:
       type: object
       required:
-        - phoneNumber
         - scopes
         - purpose
         - requestCaptureUrl
@@ -155,19 +245,9 @@ components:
         phoneNumber:
           $ref: "#/components/schemas/PhoneNumber"
         scopes:
-          type: array
-          minItems: 1
-          items:
-            type: string
-          description: |
-            List of scopes for which the Consent status is requested. The scope is a string that represents the access rights that the API Consumer is requesting from the User.
-          example:
-            - "location-verification:verify"
+          $ref: "#/components/schemas/Scopes"
         purpose:
-          type: string
-          description: |
-            The reason for which Personal Data will be processed by the API Consumer. CAMARA defines a standard set of Purposes which can be used by API Consumers to specify the reason for their intended Personal Data processing.
-          example: "dpv:FraudPreventionAndDetection"
+          $ref: "#/components/schemas/Purpose"
         requestCaptureUrl:
           type: boolean
           description: |
@@ -175,11 +255,52 @@ components:
             * `true` - If set to `true` the API will include a `captureUrl` in the response body if applicable.
             * `false` - The API will omit the consent capture URL from the response.
           example: true
+    Scopes:
+      type: array
+      minItems: 1
+      items:
+        type: string
+      description: |
+        List of scopes for which the Consent status is requested. The scope is a string that represents the access rights that the API Consumer is requesting from the User.
+      example:
+        - "location-verification:verify"      
+    Purpose:
+      type: string
+      description: |
+        The reason for which Personal Data will be processed by the API Consumer. CAMARA defines a standard set of Purposes which can be used by API Consumers to specify the reason for their intended Personal Data processing.
+      example: "dpv:FraudPreventionAndDetection"
     VerifyConsentStatusResponseBody:
       type: object
       required:
-        - consentRequired
+        - privacyStatus
       properties:
+        privacyStatus:
+          $ref: "#/components/schemas/PrivacyStatus"
+        captureUrl:
+          type: string
+          format: url
+          description: |
+            URL where the User can provide the necessary Consent. This field is only present when API Consumer requested it and if `consentRequired` is `true` and the `consentStatus` is `PENDING` or `EXPIRED`.
+          example: 'https://example.org/consent-capture-url'
+    PrivacyStatus:
+      type: array
+      description: |
+        The PrivacyStatus contains the information about the Consent status for the requested scope(s) and Purpose. It may be required more than one User Consent for the requested scope(s) and Purpose (e.g. when the requested scopes are related to multiple APIs).
+      items:
+        $ref: "#/components/schemas/PrivacyStatusObject"
+      minItems: 1
+    PrivacyStatusObject:
+      type: object
+      required:
+        - scopes
+        - purpose
+        - consentRequired
+        - consentStatus
+      properties:
+        scopes:
+          $ref: "#/components/schemas/Scopes"
+        purpose:
+          $ref: "#/components/schemas/Purpose"
         consentRequired:
           type: boolean
           description: |
@@ -198,12 +319,6 @@ components:
             - `ACTIVE`: The User has provided the necessary Consent.
             - `REVOKED`: The User has revoked the necessary Consent.
             - `EXPIRED`: The Consent provided by the User has expired.
-        captureUrl:
-          type: string
-          format: url
-          description: |
-            URL where the User can provide the necessary Consent. This field is only present when API Consumer requested it and if `consentRequired` is `true` and the `consentStatus` is `PENDING` or `EXPIRED`.
-          example: 'https://example.org/consent-capture-url'
     ErrorInfo:
       type: object
       required:
--- a/documentation/API_documentation/consent-info.yaml
+++ b/documentation/API_documentation/consent-info.yaml
@@ -1,0 +1,387 @@
+openapi: 3.0.3
+info:
+  title: Consent Info API
+  description: |
+    
+    The Consent Info API enables to verify the Consent status of a specific User to the API Consumer for the requested scope(s) and Purpose. When the Consent is required for the requested scope(s) and Purpose, but has not been provided by the User, the API also offers a mechanism for the API Consumer to request Consent from the User.
+
+    # Introduction
+
+    **TBC**
+
+    # Relevant terms and definitions
+
+    * **Consent**: An explicit opt-in action that the User takes to allow processing of Personal Data. Consent grants the API Consumer access to a set of scopes related to the User for a specific Purpose.
+    * **Purpose**: The reason for which Personal Data will be processed by an API Consumer. CAMARA defines a standard set of Purposes which can be used by API Consumers to specify the reason for their intended Personal Data processing.
+    * **TBC**: TBC
+
+    # API Functionality
+
+    This API enables an API Consumer to determine whether user Consent is required, in the applicable legislation, for the requested actions based on the provide scope(s) and Purpose. Specifically, the API:
+
+    * Determines the need for Consent: Evaluates whether user Consent is necessary according to the specified parameters and relevant legal requirements.
+    * Provides the current Consent status: If Consent is required, the API returns the current status (e.g., active, revoked, expired...) so the API Consumer can understand the current User's consent status.
+    * Offers a Consent Capture URL: If the user has not provided the necessary Consent, the API supplies an API Provider's Consent capture URL where the user can easily complete the Consent process.
+    
+    Importantly, this API does NOT delegate consent capture to the API Consumer but rather empowers the API Consumer to present the API Providers's Consent capture URL at the most opportune time and place. The actual Consent capture occurs within the API Provider's secure environment, ensuring the User's authentication with the API Provider.
+
+    **TBC**
+
+    # Authorization and authentication
+
+    The "Camara Security and Interoperability Profile" provides details of how an API consumer requests an access token. Please refer to Identity and Consent Management (https://github.com/camaraproject/IdentityAndConsentManagement/) for the released version of the profile.
+
+    The specific authorization flows to be used will be agreed upon during the onboarding process, happening between the API consumer and the API provider, taking into account the declared purpose for accessing the API, whilst also being subject to the prevailing legal framework dictated by local legislation.
+
+    In cases where personal data is processed by the API and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of three-legged access tokens is mandatory. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.
+    
+    # Further info and support
+
+    (FAQs will be added in a later version of the documentation)
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: wip
+  x-camara-commonalities: 0.5
+servers:
+  - url: '{apiRoot}/consent-info/vwip'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root, defined by the service provider, e.g. `api.example.com` or `api.example.com/somepath`
+tags:
+  - name: Verify Consent Status
+    description: Create a request to verify the Consent status
+paths: 
+  /verify:
+    post:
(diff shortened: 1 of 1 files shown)


HuubAppelboom (approved):
Looks good to me!

AxelNennker:
Does the (end-)user have to be authenticated when the API provider opens the browser with the Consent URL provided by the API?

documentation/API_documentation/consent-info.yaml
```diff
@@ -0,0 +1,387 @@
...
+paths: 
+  /verify:
+    post:
+      summary: Create a request to verify the Consent status
+      description: |
+        Create a request to verify the Consent status of a specific User to the API Consumer for the requested scope(s) and Purpose. The User is identified by the `phoneNumber` parameter in the request body. The API Consumer is identified by the `client_id` parameter deduced from the access token.
+      operationId: verifyConsentStatus
+      security:
+        - openId:
+          - consent-info:verify
```
AxelNennker, line 64: And here OIDC Authorization Code flow is started, right?

HuubAppelboom:
> Does the (end-)user have to be authenticated when the API provider opens the browser with the Consent URL provided by the API?

Yes, but I would entirely leave that up to to the Telco how to capture the consent, and whether you need to login on to that url immediately.

For example, if we are talking about a fixed line, it could also be a link to a webpage with an instruction like call number xyz and when you hear the IVR, press 1 for consent, or 2 to withdraw consent.

In the Netherlands we have for example also a website which is used by all telco's to capture consent for listing in the telephone directories (and some more stuff).  What could happen is the Consent URL is a simple instruction page explaining how to go to that website and enter the consent. That website uses partly Number Verify, partly SMS, and partly an IVR, because it is used for fixed and mobile numbers.  For some use cases it may end up here (the consent collection), for other use cases it may end up in a selfcare environment of the telco, or in an IVR, or even in a TV screen.

documentation/API_documentation/consent-info.yaml
jpengar: Not really.

jpengar: Actually it states:

> In cases where personal data is processed by the API **and users can exercise their rights through mechanisms such as opt-in and/or opt-out**, the use of three-legged access tokens is mandatory. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.

And as mentioned in https://github.com/camaraproject/IdentityAndConsentManagement/issues/224#issuecomment-2654267348, according to the assessment of Telefonica's legal team, the Consent URL API could be consumed on a "Legal Obligation" legal basis (without opt-in/opt-out) and would therefore meet the conditions agreed upon in CAMARA for an API to be consumed using 2-legged access tokens (Client Credentials grant). While a three-legged access token could also be used, user consent is not required in this case.

Moreover, the nature of the information handled by this API is essentially the same as that processed in OIDC Authorization Code Flow and CIBA. Requiring user consent here would be analogous to mandating consent for Auth Code or CIBA, which is not the case.

AxelNennker: > And as mentioned in [#224 (comment)](https://github.com/camaraproject/IdentityAndConsentManagement/issues/224#issuecomment-2654267348), according to the assessment of Telefonica's legal team, the Consent URL API could be consumed on a "Legal Obligation" legal basis (without opt-in/opt-out) and would therefore meet the conditions agreed upon in CAMARA for an API to be consumed using 2-legged access tokens (Client Credentials grant). While a three-legged access token could also be used, user consent is not required in this case.

First, would you be able to share Telefónica's privacy assessment? I assume, that that assessment has very specific preconditions that have to be met. A blanket "every API consumer can get personal information, the consent status, in all cases for all APIs seem unlikely to me.

I think that Consent URL API is very different to all other CAMARA APIs as it provides "meta" information on other CAMARA APIs.
Does every API Consumer get access to this client credentials based Consent URL/Info API?

In CAMARA usually only some API Consumers get access to client credentials based flows.
https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-credentials-flow
> The [OAuth 2.0 Client Credentials](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) grant type is used to obtain a 2-legged Access Token that does not represent a user. The grant-type can only be used if agreed between the API consumer and the API provider exposing the API, taking into account the declared purpose for accessing the API (cf. [CAMARA API Specification - Authorization and authentication common guidelines](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#camara-api-specification---authorization-and-authentication-common-guidelines)).

Getting the personal information, here whether the User did grant consent, or revoked it, should not be available to all API consumers.

If an API Consumer bought two CAMARA APIs e.g. sim-swap with purpose dpv:FraudPreventionAndDetection and device-location with purpose dpv:PersonalisedAdvertising, does the API Consumer get client credentials flow access to Consent URL API for both? 

Device-location for advertising is something for that privacy-aware user probably do not grant their consent. 
But maybe during app onboarding they checked one checkbox to quickly, but then the end-user visits the CSP's privacy transparency board and remove their consent.
Then those users would be negatively surprised if the API consumer can silently learn from the CSP that they revoked their consent and are now pestered with marketing messages to come back and grant consent again.
I think that the current consent status is personal information and any API accessing that personal information needs its own consent.

If the API consumer needs consent for Consent URL/Info API, then using OIDC Authorization Code flow in the firest place is simpler.
If the API consumer does not need consent for Consenet URL/Info API because of the purpose of the API consent status is asked for, then likely consent is not needed for that API anyway.

Additionally, introducing Consent URL API is a privacy-disaster-to-happen because of mistakes being made during API consumer onboarding.

AxelNennker:
> > Does the (end-)user have to be authenticated when the API provider opens the browser with the Consent URL provided by the API?
> 
> Yes, but I would entirely leave that up to to the Telco how to capture the consent, and whether you need to login on to that url immediately.
> 
> For example, if we are talking about a fixed line, it could also be a link to a webpage with an instruction like call number xyz and when you hear the IVR, press 1 for consent, or 2 to withdraw consent.
> 
> In the Netherlands we have for example also a website which is used by all telco's to capture consent for listing in the telephone directories (and some more stuff). What could happen is the Consent URL is a simple instruction page explaining how to go to that website and enter the consent. That website uses partly Number Verify, partly SMS, and partly an IVR, because it is used for fixed and mobile numbers. For some use cases it may end up here (the consent collection), for other use cases it may end up in a selfcare environment of the telco, or in an IVR, or even in a TV screen.

Then OIDC Authorization Code Flow just redirects to those webpages with those instructions. There is no need for a special Consent URL API because the consent url is always the same fixed one. In other words the OIDC Authorization Code Flow authentication request url _is_ the consent url.

documentation/API_documentation/consent-info.yaml
MarkCornall: I do not see how this should be client credentials. It should be 3 legged and work across aggregators, Federators and Operators. CAMARA scope is across all. Usage of 2 leg flows where user information is processed should use 3 leg flows. There will be some markets where consent will be needed.

AxelNennker:
I think that because Consent URL API is different to normal CAMARA apis it is good that we are discussing it in ICM.
Thinking about how Consent URL API might be implemented, I come up with implementation requirements that are directed at the API provider. For normal CAMARA APIs the yaml file is addressing the API consumer, while in ICM we are addressing both the API consumer and the API provider or better the Authorization Server implementation. 
The proposed API is closer to the Authorization Server, while normal APIs are closer to the resource server.

documentation/API_documentation/consent-info.yaml
HuubAppelboom: @MarkCornall @AxelNennker With this kind of reasoning you end up in an endless loop, that you have to ask user consent to see whether user consent has changed.  Also. the reasoning that three legged flow is mandatory for all cases where personal information is being processed because of the GDPR does not hold.  The GDPR does not dictate a three legged flow; for cases where user consent is required, the controller needs to ensure that informed user consent is required.  You have a similar case with for example the Telco Finder: here personal data is being processed before any user consent has been obtained, and this is done based on legitimate interest.  During onboarding of the API Consumer, you must validate that the legitimate interest is there, and that the API Consumer will not abuse the API.  That is business as usual.

AxelNennker:
I propose to split the proposed yaml file into two: consent_url.yaml and consent_info.yaml

I think the current proposed consent-info.yaml mixes several API functions into one path, which feels wrong and probably rubs against Commonalities API guidelines. 

I created https://github.com/camaraproject/IdentityAndConsentManagement/pull/271
But still think that Consent URL API is not needed because consent management can be achieve by user OIDC Authorizatin Code flow with prompt=consent.

jpengar:
> I propose to split the proposed yaml file into two: consent_url.yaml and consent_info.yaml
> 
> I think the current proposed consent-info.yaml mixes several API functions into one path, which feels wrong and probably rubs against Commonalities API guidelines.
> 
> I created #271 But still think that Consent URL API is not needed because consent management can be achieve by user OIDC Authorizatin Code flow with prompt=consent.

@AxelNennker Thank you for your detailed input. I'd like to point out that we already have an active PR and API proposal that incorporates feedback from several WG participants - including suggestions from Ericsson and Vonage. The original proposal focused solely on a Consent URL API, but based on community feedback we have broadened the scope accordingly. However, anything can be discussed. From Telefónica's point of view, it's ok to consider only the Consent URL functionality, as that was our original proposal. But it should be agreed in any case.

To keep the discussion focused and avoid fragmentation, it would be best if you could provide your feedback on this existing PR. Splitting the YAML file into two separate files and opening additional PRs could complicate the conversation and dilute the collective decision-making process.

So I would kindly ask you to refrain from opening additional PRs and keep the discussion here.

Thanks again for your input!

jpengar:
> I think the current proposed consent-info.yaml mixes several API functions into one path, which feels wrong and probably rubs against Commonalities API guidelines.

@AxelNennker Regarding your concern about mixing multiple API functions in one path, I'd like to clarify that the current design is fully compliant with the Commonalities API guidelines. The endpoint is designed with a clear, focused purpose: to provide the required consent information, along with additional fields based on the consent state. Splitting the functionality into two separate endpoints was considered, but this would require the API client to make two calls - especially when capturing consent - which is less efficient. Our goal was to streamline the process while still providing all the necessary information in a single, well-defined call.

Thanks for your feedback!

AxelNennker:
The Concent URL API proposal describes the following use cases:

> Current use cases validating the service:
> 1. Consent capture during end-user onboarding: The developer, when registering the user in a digital or physical procedure, can take advantage of the presence of the end-user during the process to pre-capture the consent for the set of APIs that the user is later going to require.
> 2. Consent capture in application notification: Developers can push a notification in their app so that customers can access directly to the consent capture process in this easy-to-access place.
> 3. Consent capture campaigns: Applications can create campaigns by email or other channels to capture the consent for substantial chunks of their user base.

re 1: this is a typical front end flow scenario. When the end-user is onboarding to API Consumer, then they create an account with API consumer and then API Consumer tells the end-user they want to use "CAMARA API" and that the end-user must consent because otherwise API-Consumer's service can't be provided. The API Consumer then redirects the end-user's browser or webview in a mobile application to OIDC Authorization Code Flow and the end-user grants consent.

re 2: this is a typical front end flow scenario: The API Consumer's mobile app receives the notification and the mobile app opens either a webview in the app to OIDC Authorization Code Flow or opens the mobile browser. In both cases, webview or mobile browser, the end-user might already be logged-in to the API provider.

re 3: this is a typical front end flow scenario: The emails contains a link to the API Consumer's campaign page. The end-user authenticates to the API Consumer and is asked for consent using OIDC ACF.

What would be an use case or scenario that cannot be done with OIDC ACF?
GDPR demands that consent is granted by a data-subject (end-user) and [Consent must be freely given, specific, informed and unambiguous](https://gdpr-info.eu/issues/consent/#:~:text=Consent%20must%20be%20freely%20given%2C%20specific%2C%20informed%20and%20unambiguous).

AxelNennker:
What changed to when you (@jpengar ) wrote in this [comment](https://github.com/camaraproject/IdentityAndConsentManagement/issues/49#issuecomment-1631037698)?

> In any case, so far there is no need to expose a Consent API to third party clients as a CAMARA API with CRUD operations on consent records. Consent must be obtained through operator channels (there is no delegated consent) and only operators internally manage user consents (e.g. check if a consent record exists, create a new consent record if it is captured from the user, etc).

Consent must still be obtained through operator channels - We are using OIDC Authorization Code Flow or CIBA for that.

The flow is determined at API Consumer Onboarding time. In some cases consent might never be needed and two-legged flow might be legal. CAMARA then uses OIDC ACF to identify the subscriber and if the subscriber can reasonably be equated with the end-user determines that consent does not need to be collected during OIDC ACF. CAMARA thus uses OIDC ACF basically as a two-legged flow.
I think using CIBA in a similar manner, is probably **not** possible without sending a message. Please see [CIBA privacy](https://github.com/camaraproject/IdentityAndConsentManagement/issues/258).

jpengar:
> What changed to when you (@jpengar ) wrote in this [comment](https://github.com/camaraproject/IdentityAndConsentManagement/issues/49#issuecomment-1631037698)?
> 
> > In any case, so far there is no need to expose a Consent API to third party clients as a CAMARA API with CRUD operations on consent records. Consent must be obtained through operator channels (there is no delegated consent) and only operators internally manage user consents (e.g. check if a consent record exists, create a new consent record if it is captured from the user, etc).
> 
> Consent must still be obtained through operator channels - We are using OIDC Authorization Code Flow or CIBA for that.
> 
> The flow is determined at API Consumer Onboarding time. In some cases consent might never be needed and two-legged flow might be legal. CAMARA then uses OIDC ACF to identify the subscriber and if the subscriber can reasonably be equated with the end-user determines that consent does not need to be collected during OIDC ACF. CAMARA thus uses OIDC ACF basically as a two-legged flow. I think using CIBA in a similar manner, is probably **not** possible without sending a message. Please see [CIBA privacy](https://github.com/camaraproject/IdentityAndConsentManagement/issues/258).

@AxelNennker There is nothing in Telefónica's Consent URL proposal that contradicts the statement you just copied above completely out of context by the way (which is a comment on a question from 2023 about having a CRUD Consent Management API that would allow third parties to manage consents on behalf of the API provider and thus delegate consent management, which is NOT what is being proposed here).

Please take a look at PR description:
https://github.com/camaraproject/IdentityAndConsentManagement/pull/266#issue-2845976385
> Importantly, this API does NOT delegate consent capture to a third party but rather empowers the API Consumer to present the API Provider's consent capture URL at the most opportune time and place. The actual consent capture occurs within the API Provider's secure environment, ensuring the user's authentication with the API Provider.

The only thing added in this PR to Telefónica's original proposal (https://github.com/camaraproject/APIBacklog/issues/68) is the ability to get the consent status (which is far from being a CRUD API and which is NOT a consent management API, not a consent delegation API). It has been added taking into account feedback from other participants, and of course for discussion. As said before, https://github.com/camaraproject/IdentityAndConsentManagement/pull/266#issuecomment-2662413886, from Telefónica's point of view it's ok to consider only the Consent URL functionality, as that was our original proposal. But it should be agreed in any case.

jpengar:
> I wonder whether the phoneNumber parameter should be optional then and maybe omitted.

No. The API consumer needs to identify the user whose consent is to be captured with the obtained consent capture URL.

jpengar:
> The flow is determined at API Consumer Onboarding time. In some cases consent might never be needed and two-legged flow might be legal. CAMARA then uses OIDC ACF to identify the subscriber and if the subscriber can reasonably be equated with the end-user determines that consent does not need to be collected during OIDC ACF. CAMARA thus uses OIDC ACF basically as a two-legged flow.

Please, read carefully again the agreed CAMRA statement because it does not say the same thing

> In cases where personal data is processed by the API and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of three-legged access tokens is mandatory. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.

And to be clear, I'm NOT mandating or requiring this API to be used specifically consumed in two-legged mode. What I have said here, https://github.com/camaraproject/IdentityAndConsentManagement/pull/266#discussion_r1954158076, is that according to the assessment of Telefonica's legal team, the Consent URL API could be consumed on a "Legal Obligation" legal basis (without opt-in/opt-out) and would therefore meet the conditions agreed in CAMARA for an API to be consumed using 2-legged access tokens (client credentials grant). **While a 3-legged access token could also be used, user consent is not required in this case.**. Period, nothing more.

That's why the API DRAFT proposal so far sticks to the `info.description` template defined by ICM.

AxelNennker:
You clarified that Consent URL API is not about delegated consent. I do not doubt that.

I still do not see use case or scenarios where Consent URL API helps. 
For [weeks](https://github.com/camaraproject/APIBacklog/pull/67) I proposed use case and scenarios where Consent URL might help which in the end could all be done by OIDC ACF.

@jpengar wrote:
> The key benefit of the Consent URL API is that it offers a significant advantage by enabling a fully decoupled, out-of-band mechanism for consent capture. This API allows the application to control the user experience entirely. The application can generate a consent URL from the Operator, which the user must authenticate with. However, the application has the flexibility to deliver this URL to the user through the most appropriate channel or medium, tailored to the user's current context.

Why are "fully decoupled" or out-of-band mechanisms a benefit? For the end-user to grant consent, there must be interaction with the end-user. The end-user is on a browser or mobile application and OIDC ACF can be used?

Please provide a use case where OIDC ACF is inferior to Consent URL API.
Claiming that some is better does not make it so. Please explain.

HuubAppelboom:
A use case where you can not use OIDC ACF is for where you have terminals that don't have a web interface, for example a fixed line phone.  ACF can only work if you have a selfcare environment for that phone, but that is not always the case / users may not be using that.

AxelNennker:
Is the reason for proposing a new Consent API that current MNO implementations insist on network-based authentication and some MNOs did not implement asking for consent? 

OIDC Authorization Code Flow was designed to be an authentication and consent protocol, why not use it?

AxelNennker:
> > I wonder whether the phoneNumber parameter should be optional then and maybe omitted.
> 
> No. The API consumer needs to identify the user whose consent is to be captured with the obtained consent capture URL.

Why? The API provider must authenticate the User whose consent is to be captured.

The Consent URL API must not reveal personal information about the User and the best way to ensure that is that the Consent URL does not depend on the User at all.

If e.g. the consent capture url returned by Consent URL API would be https://ageover80.mno.com/authorize?login_hint=tel:+123456789 then the Consent URL API leaked private data. Better Consent URL API does not know the phone number.
And if the end-user is later authenticated anyway at capture URL then we could have use OIDF Authentication Code Flow from the beginning.

jpengar:
> > > I wonder whether the phoneNumber parameter should be optional then and maybe omitted.
> > 
> > 
> > No. The API consumer needs to identify the user whose consent is to be captured with the obtained consent capture URL.
> 
> Why? The API provider must authenticate the User whose consent is to be captured.
> 
> The Consent URL API must not reveal personal information about the User and the best way to ensure that is that the Consent URL does not depend on the User at all.
> 
> If e.g. the consent capture url returned by Consent URL API would be https://ageover80.mno.com/authorize?login_hint=tel:+123456789 then the Consent URL API leaked private data. Better Consent URL API does not know the phone number. And if the end-user is later authenticated anyway at capture URL then we could have use OIDF Authentication Code Flow from the beginning.

The API consumer must provide the API provider with the identity of the user for whom consent is required. The API provider will then generate a consent URL for that combination of user identity, purpose and scope(s). I don't know why you assume a consent URL format like `https://ageover80.mno.com/authorize?login_hint=tel:+123456789`... In any case, each API provider could apply their own security considerations and mechanisms to the consent URL according to their requirements. But in Telefónica (say), for example, we could use a temporary and non-predictable URL that could look like `https://ob-url.com?jwt=xxx...yyy...zzz`, which does not provide any user information.

documentation/API_documentation/consent-info.yaml
```diff
@@ -0,0 +1,387 @@
...
+          Consent status verification request
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/VerifyConsentStatusRequestBody"
+            example:
+              phoneNumber: "+123456789"
+              scopes:
+                - "location-verification:verify"
+              purpose: "dpv:FraudPreventionAndDetection"
```
chrishowell, line 81: Is there a reason why we couldn't allow a list of scope/purposes on this API? It would be nice to be able to have a single call to obtain all of the statuses at once, and/or receive the consent URL to collect consent for all of those scope/purposes?

chrishowell: I think this comment can be discussed when we review the actual API spec, rather than whether the API should exist; I am/Vonage is in agreement that this API is required and is much more suitable/scaleable than using Auth Code (prompt=none) for this purpose.

AxelNennker:
@jpengar 
> The API consumer must provide the API provider with the identity of the user for whom consent is required. The API provider will then generate a consent URL for that combination of user identity, purpose and scope(s). I don't know why you assume a consent URL format like `https://ageover80.mno.com/authorize?login_hint=tel:+123456789`... In any case, each API provider could apply their own security considerations and mechanisms to the consent URL according to their requirements. But in Telefónica (say), for example, we could use a temporary and non-predictable URL that could look like `https://ob-url.com?jwt=xxx...yyy...zzz`, which does not provide any user information.

I am not assuming that the capture URL looks like `https://ageover80.mno.com/...` That is an example how Consent URL API could misused or (accidentally) implemented wrong. That is an example where the capture URL is leaking information about the user. That is only possible if the user identifier is part of the API request. If the API request does not contain an user identifier then this cannot happen.

Still, (again) what is the use case for Consent Capture URL what cannot be done with OIDC Authorization Code flow?

AxelNennker:
> A use case where you can not use OIDC ACF is for where you have terminals that don't have a web interface, for example a fixed line phone. ACF can only work if you have a selfcare environment for that phone, but that is not always the case / users may not be using that.

At some point in time the user of the fixed line phone is using the Internet - with or without a Consent URL API.
That is the time when OIDC Authorization Code flow is used. That is the time when the user authenticates using their user authentication credentials the CSP provided. That is the normal OpenId Connect flow.

Even if some insist on network-based authentication for OIDC ACF. 
The Authorization Server is there to help the user and to enable the API Consumer to use our APIs.
The API Provider and the Authorization Server should always allow the end-user to authenticate and provide their consent.

AxelNennker:
> A use case where you can not use OIDC ACF is for where you have terminals that don't have a web interface, for example a fixed line phone. ACF can only work if you have a selfcare environment for that phone, but that is not always the case / users may not be using that.

At some point in time the user of the fixed line phone is using the Internet - with or without a Consent URL API.
That is the time when OIDC Authorization Code flow is used. That is the time when the user authenticates using their user authentication credentials the CSP provided. That is the normal OpenId Connect flow.

Even if some insist on network-based authentication for OIDC ACF. 
The Authorization Server is there to help the user and to enable the API Consumer to use our APIs.
The API Provider and the Authorization Server should always allow the end-user to authenticate and provide their consent.

HuubAppelboom:
> > A use case where you can not use OIDC ACF is for where you have terminals that don't have a web interface, for example a fixed line phone. ACF can only work if you have a selfcare environment for that phone, but that is not always the case / users may not be using that.
> 
> At some point in time the user of the fixed line phone is using the Internet - with or without a Consent URL API. That is the time when OIDC Authorization Code flow is used. That is the time when the user authenticates using their user authentication credentials the CSP provided. That is the normal OpenId Connect flow.
> 
> Even if some insist on network-based authentication for OIDC ACF. The Authorization Server is there to help the user and to enable the API Consumer to use our APIs. The API Provider and the Authorization Server should always allow the end-user to authenticate and provide their consent.

@AxelNennker The more I think about this, I still do not see any reason why a web interface is the only interface through which user consent can be acquired, and why CAMARA should restrict itself to consent acquisition through a web interface, especially when we are talking about API's which are normally run in the background with the CIBA protocol.

For example, in case you need to acquire consent for example for a Scam Signal API, you may want to do this through about any interface by why end users interact with the Telco. An online selfcare environment can be very important for that, but if you have a goal to reach 99% of the population, you may want to use other channels as well in which users are alos authenticated.  Just to name a few:
- when the customer is visiting a shop for a servicequestion, the sales rep may alsoask them whether want to have it turned on. I don't think it is very customer friendly to aks customersto login through a web interface in the shop, when you ahve already authenticated them by their ID card
- on the phone, when a customer has been autneticated, you can also ask for permission to switch a service on
- interaction by SMS; this can be very useful if you have a promotion for a servicethrough main stream mass media (TV, Radio, newspapers, or when you have an SMS campaign
- same for a campaign through chat channels / RCS
- turning a service on by using an IVR
- or activating a service by short code on the phone, similar to call diversion

All these interactions without using a web interface, and are usually very instrumental in getting conversion up.  We should support any interface for consent collection (as long as the user is sufficiently authenticated).  Also to my experience, lost of users have lost their selfcare password a long time ago, and are not daily active in the selfacre portalof their telco.  If you want to be successfull in consent collection, you should also include channels which are easy and relevant for users.  So, can you explain why you think a web interface for consent collection is a must ???

jpengar:
FYI, these are the updates in https://github.com/camaraproject/IdentityAndConsentManagement/pull/266/commits/216fa03eea445466ffc099fdc77b2b01b200eaee:

- Identifying the phone number from the access token: Updated the info.description section to align with CAMARA Commonalities guidelines, as done in other CAMARA APIs. The API can be consumed with both two-legged and three-legged access tokens, making phoneNumber optional in the request.
- Response body updates: Modified to accommodate multiple consents. For instance, if the scope list corresponds to multiple APIs, more than one user consent may be required.
- Updated examples: Refined request and response body examples to provide better clarity.

MarkCornall:
As previously said I do not support this being deployed as a 2-legged flow. It is personal information. You are asking about the status of an API that requires consent (or you wouldn't call the API) you retrieve information on consent status. This all points to a 3-legged flow being necessary.

HuubAppelboom:
@MarkCornall Hi Mark, for many use cases (like an anti-fraud check during a transaction), 3 legged flow is much too slow. that will not work.  

In any case, both the API Consumer and the API Provider should be aware of the consent status for that particular user in combination with combination with the API Cosnumer's application.  Why would it be needed to be kept a secret what the consent status is for that API Consumer ??

AxelNennker:
@HuubAppelboom 
> All these interactions without using a web interface, and are usually very instrumental in getting conversion up. We should support any interface for consent collection (as long as the user is sufficiently authenticated). Also to my experience, lost of users have lost their selfcare password a long time ago, and are not daily active in the selfacre portalof their telco. If you want to be successfull in consent collection, you should also include channels which are easy and relevant for users. So, can you explain why you think a web interface for consent collection is a must ???

I never said that a web interface "is a must". A web interface is just what, I think, _everybody_ is using. If a telco has a better way then go for it.

I never insisted that consent capture must be "in-flow". It is just that OpenID is THE (web) protocol to trigger User authentication and consent capture. Every API Provider who uses a standard, hopefully [certified](https://openid.net/certification/#OPENID-OP-P),  OpenId Provider supports that the end-user is authenticated and consent collection is started. Afterwards the API Consumer gets a code and from that code the API Consumer gets an access token and refresh token and an id_token. Those can be used at any time in OIDC Authorization Code Flow and CIBA.

I am recommending to use existing standards. 
I am recommending that if the API Consumer have their own mobile application that that mobile application uses standard protocols.
I am recommending that if the end-user is on a mobile device that then web-standards are used.

If there is a standard way in any other protocols to collect consent, then I am all supporting that.

If the API Provider has a Web-Portal for end-user consent management, then I am all supporting that. Usually "modern" Web-Portals have Web-Part and a Backend-Part. Integration cost are much lower if standards are used - for the front-end, Web-Part and for the Backend-Part. It seems natural, for me, to use OIDC Authorization Code Flow for any front-end flow.
If the front-end needs a client_id, scopes etc, then those should be the same as defined in the standard OIDC ACF.

The end-user can also direct their browser to some consent management URL at the API provider. Then the API provider can guide the end-user to grant consent for `purpose` and scopes to client_id/application_ids. 

If the API Provider e.g. decides to allow service center agents, to manage the consent for their web-averse end-users, then that is a business decision. But this is not what ICM is about.

In CAMARA we decided that:
> CAMARA API access will be secured using [OpenID Connect](https://openid.net/specs/openid-connect-core-1_0.html) (OIDC) on top of [OAuth 2.0 protocol](https://datatracker.ietf.org/doc/html/rfc6749) following the [CAMARA Security and Interoperability Profile](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md).

The API Provider can do whatever they want during consent management, I do not know of a widely used standard for consent management. There was some work done around that topic e.g. [Kantara's Consent Receipt](file:///Users/axelnennker/Downloads/consent-receipt-specification-v1-1-0.pdf) work.

I do not think and never thought that a web interface for Consent Collection is a must.
I do think that using a standard is in the API Providers' own best interest.

I think it is natural and saving the API providers resources, if the API Provider use their already implemented OpenId Provider to trigger consent capture/management.

HuubAppelboom:
@AxelNennker In which jurisdiction is a three legged flow mandatory for the processing of personal data when the legal basis is legitimate interest ??  I don't know of any law in the EU that prescribe you must use a three legged flow...

AxelNennker:
Move discussion to https://github.com/camaraproject/ConsentInfo/pull/4

jpengar:
As per https://github.com/camaraproject/IdentityAndConsentManagement/issues/278#issuecomment-2728951241, I close this PR.

> @AxelNennker Once the API and this new repository creation was approved by [ICM](https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/83689619/2025-02-26+ICM+Draft+Agenda+Minutes) and [TSC](https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/83199399/2025-03-06+TSC+Minutes) the new repository has been created https://github.com/camaraproject/ConsentInfo (Thanks @hdamker)
> 
> I've added https://github.com/camaraproject/ConsentInfo/issues/3 to create the first API draft version to continue the API design and use case discussion. I' ve created the corresponding PR https://github.com/camaraproject/ConsentInfo/pull/4 with the same content as https://github.com/camaraproject/IdentityAndConsentManagement/pull/266.
> 
> All the context has been linked to the issue and the PR, but I think it's cleaner to start a "new" discussion in the repository and link all the context for further information:
> 
> https://github.com/camaraproject/APIBacklog/pull/67
> https://github.com/camaraproject/APIBacklog/issues/68
> https://github.com/camaraproject/IdentityAndConsentManagement/issues/224
> https://github.com/camaraproject/IdentityAndConsentManagement/pull/266
> https://github.com/camaraproject/IdentityAndConsentManagement/pull/271
> So I propose to close https://github.com/camaraproject/IdentityAndConsentManagement/issues/224, https://github.com/camaraproject/IdentityAndConsentManagement/pull/266 and https://github.com/camaraproject/IdentityAndConsentManagement/pull/271 with an appropriate comment pointing to the new repository in accordance with the latest decisions made in ICM and TSC.

PR #285, opened 2025-03-27
mandatory fields in signed ACF request objects

### LINKED ISSUE #283: Claims in signed authentication requests
**Problem description**

In the [Signed Authentication Requests section](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#signed-authentication-requests) of the Authorization Code flow, there is no mention of which typical claims are mandatory, as is already the case in the [CIBA specification](https://openid.net/specs/openid-client-initiated-backchannel-authentication-core-1_0.html#rfc.section.7.1.1).

**Expected action**

To mitigate possible attacks, as we commented in a [previous issue](https://github.com/camaraproject/IdentityAndConsentManagement/issues/128#issuecomment-1968294433), we should make the following claims mandatory: `iss`, `aud`, `iat`, `exp` and `jti`.
 
In addition, we should apply the same maximum request object lifetime policy in both CIBA and Authorization Code flows that we require in the [client assertion](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-authentication) for authentication.

**Additional context**

AxelNennker:
#### What type of PR is this?

* security
* enhancement/feature

OIDC recommends some fields in signed request objects.
This PR make some required.

Added April 12, 2025: Restrict the lifetime of signed request objects to 300 seconds for both Authorization Code Flow and CIBA. Removed "ACF" from the title, so that PR is not restricted to ACF because lifetime of signed request objects is now max 300s for both ACF and CIBA.

Fixes #283 

#### Changelog input

```
 For improved security some fields in signed request objects in OIDC Authorization Code Flow are now mandatory.

```

### DIFF
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -116,7 +116,7 @@ It is RECOMMENDED that signed authentication requests be used, as specified by [
   * Values for the mandatory parameters `scope`, `response_type`, `client_id` and `redirect_uri` MUST be included using the OAuth2.0 request syntax. The values for these parameters MUST match those in the signed request object.
   * The fields iss, aud, iat, exp and jti MUST be included in the signed request object.
   * For all other parameters, the authorization server MUST only use the parameters contained in the signed request object. 
-  * The request SHALL be rejected by the authorisation server if the exp claim is more than 300 seconds later than the time of receipt. Additionally, if the iat claim is present, the request SHALL be rejected if the difference between the exp claim and iat claim is more than 300 seconds.
+  * The request SHALL be rejected by the authorisation server if the iat claim is later than the time of receipt, or if the difference between the exp claim and iat claim is greater than 300 seconds.
 
 It is RECOMMENDED that the value of the `aud` field of the signed authentication request is the URL of the [Authorization Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint).
 The authorization server MAY accept different values of the `aud` field e.g. the `issuer` field of its [OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata). 
--- a/documentation/CAMARA-Security-Interoperability.md
+++ b/documentation/CAMARA-Security-Interoperability.md
@@ -114,7 +114,9 @@ CAMARA does not intend to limit the Authorization Code Flow to just network-base
 
 It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication). In addition, this document defines that:
   * Values for the mandatory parameters `scope`, `response_type`, `client_id` and `redirect_uri` MUST be included using the OAuth2.0 request syntax. The values for these parameters MUST match those in the signed request object.
+  * The fields iss, aud, iat, exp and jti MUST be included in the signed request object.
   * For all other parameters, the authorization server MUST only use the parameters contained in the signed request object. 
+  * The request SHALL be rejected by the authorisation server if the exp claim is more than 300 seconds later than the time of receipt. Additionally, if the iat claim is present, the request SHALL be rejected if the difference between the exp claim and iat claim is more than 300 seconds.
 
 It is RECOMMENDED that the value of the `aud` field of the signed authentication request is the URL of the [Authorization Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint).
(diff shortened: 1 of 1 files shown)


documentation/CAMARA-Security-Interoperability.md
```diff
@@ -114,7 +114,9 @@ CAMARA does not intend to limit the Authorization Code Flow to just network-base
 
 It is RECOMMENDED that signed authentication requests be used, as specified by [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests). The same key MAY be used for signing the authentication request as is used for [client authentication](#client-authentication). In addition, this document defines that:
   * Values for the mandatory parameters `scope`, `response_type`, `client_id` and `redirect_uri` MUST be included using the OAuth2.0 request syntax. The values for these parameters MUST match those in the signed request object.
+  * The fields iss, aud, iat, exp and jti MUST be included in the signed request object.
   * For all other parameters, the authorization server MUST only use the parameters contained in the signed request object. 
+  * The request SHALL be rejected by the authorisation server if the exp claim is more than 300 seconds later than the time of receipt. Additionally, if the iat claim is present, the request SHALL be rejected if the difference between the exp claim and iat claim is more than 300 seconds.
```
garciasolero, line 119: @AxelNennker, I would add the same paragraph regarding the expiration date in the CIBA flow section.

AxelNennker: Maybe I am confused, this is the text from the CIBA section:
> The request SHALL be rejected by the authorisation server if the exp claim is more than 300 seconds later than the time of receipt. Additionally, if the iat claim is present, the request SHALL be rejected if the difference between the exp claim and iat claim is more than 300 seconds.

It is exactly the same.

sebdewet:
[sad]   DEWET Sébastien INNOV/IT-S reacted to your message:
________________________________

sebdewet:
[0]     DEWET Sébastien INNOV/IT-S reacted to your message:
________________________________

documentation/CAMARA-Security-Interoperability.md
jpengar: @AxelNennker The text quoted above is actually from the [Client Authentication](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-authentication) section, not the [CIBA](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-initiated-backchannel-authentication-flow) section. And I think what @garciasolero meant above is that the quoted text should also be added to the CIBA section in https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#authentication-request.

jpengar: This would fix the part of issue #283 below:
> In addition, we should apply the same maximum request object lifetime policy in both CIBA and Authorization Code flows that we require in the [client assertion](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-Security-Interoperability.md#client-authentication) for authentication.

garciasolero: In the current profile included in this pull request, there are only two sections addressing the JWT lifetime limited to 300 seconds: 

1. in the paragraph added in this PR for the request parameter within the auth code flow
2. in the section of the client assertion for authentication that already existed in previous versions of the profile. 

There is no mention of the JWT lifetime for the request parameter in the CIBA section. If we limit it for those cases, we should also limit it for CIBA as proposed in the issue.

AxelNennker:
Maybe we should add a lifetime restriction to all signed request objects by moving the new restriction on lifetimes outside the flow sections?

Please review again.

garciasolero (approved):

jpengar (dismissed):
LGTM

documentation/CAMARA-Security-Interoperability.md
eric-murray, line 119: > if the `iat` claim is present

But the `iat` claim MUST be present, because it is mandatory. So the text can be simplified.

```suggestion
  * The request SHALL be rejected by the authorisation server if the iat claim is later than the time of receipt, or if the difference between the exp claim and iat claim is greater than 300 seconds.
```

AxelNennker:
I resolved the merge conflicts. Please review again and consider approving.

documentation/CAMARA-Security-Interoperability.md
jpengar: I've committed this change to merge the PR, as it has already been considered in other related discussions.

jpengar (approved):
LGTM after adding Eric's suggestion

PR #287, opened 2025-04-08
Update request object errors

garciasolero:
#### What type of PR is this?

* documentation

Although the use of authorization requests signed with a request object in the Authorization Code and CIBA flows is recommended on the client side, their implementation is mandatory on the authorization server side. 

Therefore, the `request_not_supported` error will never be returned to the application. Additionally, the `invalid_request_object` error must be added in case the application sends this parameter incorrectly.


jpengar (dismissed):
LGTM

sebdewet (dismissed):
LGTM

AxelNennker (changes requested):
Even if support is mandatory we should keep the error, I think.

Hopefully, certification and test suites catch the case were an CAMARA AZ does not support signed request and thus this error should, indeed, never happen. 

Does the GSMA certification / test-suite have test cases with signed request objects?

Should we clarify that support for signed objects is mandatory for API Providers?

AxelNennker:
I suggest to add a requirement that API Providers MUST support signed requests.
Then I would remove request_not_supported and not add the new error but just return BAD REQUEST

garciasolero:
> I suggest to add a requirement that API Providers MUST support signed requests. Then I would remove request_not_supported and not add the new error but just return BAD REQUEST

Added requirement to support signed requests as mandatory.

AxelNennker (approved):

jpengar (approved):
LGTM.

PR #291, opened 2025-04-14
Add JWT Bearer Token Flow to  CAMARA-API-access-and-user-consent.md

### LINKED ISSUE #277: Create OAuth2 two-legged flow with user identifier in access token
**Problem description**

**Possible evolution**

**Alternative solution**

**Additional context**

AxelNennker:
#### What type of PR is this?

* enhancement/feature

For legitimate-use use cases we need a two-legged flow that creates an access token that is associated with a user

Fixes #277


AxelNennker:
I am moving this branch' changes to 290. 
I should have done this from the start, sorry.

PR #293, opened 2025-05-27
Fall25 task force - CIBA privacy fix - TEF DRAFT proposal

### LINKED ISSUE #258: CIBA Privacy

jpengar:
#### What type of PR is this?

* documentation

This is Telefónica's proposal to fix CIBA Privacy Issue #258, based on the discussions held in #258 and #268. This PR is intended solely to provide a clean draft proposal to facilitate discussions for the Fall25 task force created in the ICM WG.

Fixes #258 

The PR was created purely for the sake of the task force, to facilitate reaching an agreement on a clean proposal that we might end up agreeing on.

CC @subha5h

#### Changelog input

```
CIBA auth and privacy clarifications
```

#### Additional documentation 

* https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/123077145/2025-05-21+Draft+Agenda+ICM+Minutes
* https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/126878096/2025-05-27+Fall25+task+force+1+-+Draft+Agenda+ICM+Minutes

### DIFF
--- a/documentation/CAMARA-API-access-and-user-consent.md
+++ b/documentation/CAMARA-API-access-and-user-consent.md
@@ -213,12 +213,8 @@ alt OIDC Client-Initiated Backchannel Authentication (CIBA) Flow between API Con
   opt If User Consent is required for the legal basis of the purpose  
     ExpO->>Consent: Check if Consent is granted
   end
-  alt If Consent is Granted or Consent not needed for legal basis   
-    ExpO->>BE: HTTP 200 OK {"auth_req_id": "{OperatorAuthReqId}"  
-  else If Consent is needed and is NOT granted - Out Of Band Consent Capture (Push/SMS/other)
-    Note over ExpO,User: User Interaction <br> out-of-band capture consent mechanism chosen by the Operator
-    ExpO->>BE: HTTP 200 OK {"auth_req_id": "{OperatorAuthReqId}"}
-  end
+  Note over ExpO,User: Start User authentication and Request authorization process following<br>Section 8 of the CIBA Core 1.0 spec. Operator MUST obtain an authorization<br> decision as described in Section 3.1.2.4 of the OIDC Core 1.0 spec.<br>This MUST include a mechanism to capture User Consent if required.
+  ExpO->>BE: HTTP 200 OK {"auth_req_id": "{OperatorAuthReqId}"}
   loop API Consumer polls until consent is granted or until expires. If granted in advance, token returned in first poll
     BE->>+ExpO: POST /token <br>Credentials}<br>auth_req_id={OperatorAuthReqId}    
     ExpO->>-BE: HTTP 200 OK <br>{"access_token": "{OperatorAccessToken}"}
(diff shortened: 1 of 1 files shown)


AxelNennker:
This flow always returns an auth_request_id regardless whether the end-user already granted their consent, right?
As long as the ip address or msisdn belong to the user.

That would then not leak PII, I think.

AxelNennker:
if an identifer like MSISDN or ipport is used with this CIBA flow, then the flow is a two-legged flow because the User is not involved. A prio involvement by e.g. having given consent out-of-band beforehand does not solve this being a two-legged flow because the user has not control over the authentication request parameters like login_hint.

The attacker enters the victims MSISDN and the API Provider leaks the PII whether the end-user gave their consent or not.

shilpa-padgaonkar (approved):
https://github.com/camaraproject/IdentityAndConsentManagement/pull/293#issuecomment-2915739841

LGTM

jpengar:
> @jpengar @AxelNennker
> 
> Following yesterday’s taskforce call, I wanted to share my thoughts on the topic. I see this version as an improvement over the previous text. We interpret this new text in a way that in CIBA, User authentication is mandatory while consent capture is not mandatory as the user might have already granted consent and it might still be valid.
> 
> Now our primary focus be on finalizing the JWT bearer token flow, enabling use cases that don't require user interaction to transition from CIBA to the JWT bearer token flow ASAP.

If by "User authentication is mandatory while consent capture is not mandatory as the user might have already granted consent and it might still be valid." you mean that user interaction is mandatory in CIBA and that JWT-Bearer would enable scenarios without user interaction, that is not what the text says.

That is also not aligned with existing implementations until [Spring 25](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#ciba-flow-backend-flow).

The text essentially maintains backwards compatibility, but also gives API providers the freedom to choose how to authorise requests, which may or may not involve user interaction depending on the use case and always in accordance with applicable law. An API provider could now always contact the user (if applicable legislation requires so), but also they may rely in the login_hint according to their own risk/legal assessment for example because the API Consumer (say a bank) is in under a contract relationship where the API Consumer is responsible to validate those MSISDNs and to send requests only for MSISDNs that corresponds to customers of the API Consumer.

**In any case, we agree completely that this text represents the best way forward, and we are willing to merge it as is. However, it is crucial that we are aligned in our interpretation of it.**

sebdewet (approved):
LGTM

